PR #23, opened 2019-08-28
(FACT-1999) Add windows facts

oanatmaria:


lib/resolvers/windows/dmi_bios_resolver.rb
```diff
@@ -0,0 +1,32 @@
...
+    @@semaphore = Mutex.new
+    @@fact_list ||= {}
+
+    def resolve(fact_name)
+      @@semaphore.synchronize do
+        result ||= @@fact_list[fact_name]
+
+        return result if result
+
+        win = Win32Ole.new
```
gimmyxd: You can extract the bellow code as an `private` method:
```ruby
class DMIBiosResolver < BaseResolver
  class << self
    # Manufacturer
    # SerialNumber
    @@semaphore = Mutex.new
    @@fact_list ||= {}

    def resolve(fact_name)
      @@semaphore.synchronize do
        result ||= @@fact_list[fact_name]
        result || read_fact_from_bios(fact_name)
      end
    end

    private

    def read_fact_from_bios(fact_name)
      win = Win32Ole.new

      bios = win.exec_query('SELECT Manufacturer,SerialNumber from Win32_BIOS').to_enum.first

      build_fact_list(bios)

      @@fact_list[fact_name]
    end

    def build_fact_list(bios)
      @@fact_list[:manufacturer] = bios.Manufacturer
      @@fact_list[:serial_number] = bios.SerialNumber
    end
  end
end
```

lib/resolvers/windows/processors_resolver.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    def find_isa(arch)
+      architecture_hash =
+        { 0 => 'x86', 1 => 'MIPS', 2 => 'Alpha', 3 => 'PowerPC', 5 => 'ARM', 6 => 'Itanium', 9 => 'x64' }
+      isa = architecture_hash[arch]
+
+      return isa if isa
+
+      raise 'Unable to determine processor type: unknown architecture'
```
gimmyxd: Should Facter raise specific errors for different situations?
This will raise RuntimeError which is very broad.
cc: @IrimieBogdan

lib/resolvers/windows/processors_resolver.rb
```diff
@@ -0,0 +1,69 @@
...
+        proc = win.exec_query('SELECT Name,Architecture,NumberOfLogicalProcessors FROM Win32_Processor')
+
+        result = iterate_proc(proc)
+        build_fact_list(result)
+
+        @@fact_list[fact_name]
+      end
+    end
+
+    def invalidate_cache
```
sebastian-miclea: It would be easier to have this in the BaseResolver and have it available everywhere.

IrimieBogdan:
Branch name should be the same as task name (e.g. FACT-1999 instead of WinFacts)

lib/resolvers/windows/processors_resolver.rb
IrimieBogdan: We should have specific errors.

oanatmaria: Changed it to debug log

spec/facter/resolvers/windows/processors_resolver_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    before do
+      win = double('Win32Ole')
+      processor = double('proc', Name: 'Pretty_Name', Architecture: 10, NumberOfLogicalProcessors: 0)
+
+      allow(Win32Ole).to receive(:new).and_return(win)
+      allow(win).to receive(:exec_query).with('SELECT Name,Architecture,NumberOfLogicalProcessors FROM Win32_Processor')
+                                        .and_return([processor])
+    end
+
+    it 'should raise error' do
```
gimmyxd: this is not raising, is just logging.
Also not sure if you are using a guideline  for writing specs but might be useful:
http://www.betterspecs.org/#should

spec/facter/resolvers/windows/dmi_bios_resolver_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+describe 'Windows DMIBiosResolver' do
+  before do
+    win = double('Win32Ole')
```
gimmyxd: you could extracts this in `let` so the value can be changed in different contexts, without the need to redefine the `allow(Win32Ole).to receive(:new).and_return(win)`

http://www.betterspecs.org/#let

PR #27, opened 2019-09-02
(FACT-2000) Add support for legacy facts.

BogdanIrimie:


lib/utils/query_parser.rb
```diff
@@ -67,7 +70,11 @@ def self.construct_loaded_fact(query_tokens, query_token_range, fact_name, klass
       filter_tokens = query_tokens - query_tokens[query_token_range]
 
       user_query = @uq.any? ? query_tokens[query_token_range].join('.') : ''
-      SearchedFact.new(fact_name, klass_name, filter_tokens, user_query)
+      if fact_name.end_with? '.*'
+        SearchedFact.new(user_query, klass_name, filter_tokens, user_query)
```
sebastian-miclea: I think this would look better if we move it to the SearchedFact class and just override the getter for @name.
Something like:
`  def name
      @name.end_with? '.*' ? @user_query : @name
    end`

It's a bit confusing to see the same param been sent twice.

lib/utils/fact_loader.rb
```diff
@@ -21,7 +21,8 @@ def self.load(operating_system)
 
       classes.each do |class_name|
         klass = Class.const_get("Facter::#{os}::" + class_name.to_s)
-        fact_name = klass::FACT_NAME
+        # fact_name = klass::FACT_NAME
+        fact_name = klass::FACT_NAME.end_with?('.*') ? klass::FACT_NAME.gsub('.*', 'regexfact') : klass::FACT_NAME
```
sebastian-miclea: I think this can be simplified to:
`fact_name = klass::FACT_NAME.gsub('.*', 'regexfact')`

lib/facts/macosx/network_ip.rb
```diff
@@ -12,7 +12,7 @@ def initialize(*args)
       end
 
       def call_the_resolver
-        Fact.new(FACT_NAME, 'l92.l68.O.l')
+        [Fact.new(FACT_NAME, 'l92.l68.O.l')]
```
sebastian-miclea: Isn't there a way we could avoid having all facts return an array even if there is only one element?

lib/facter.rb
```diff
@@ -20,51 +20,78 @@ def self.value(*args)
...
-    def enrich_searched_fact_with_value!(searched_facts, fact)
-      matched_fact = searched_facts.select { |elem| elem.name == fact.name }
-      matched_fact.first.value = fact.value
+    def enrich_searched_fact_with_values(searched_facts, facts)
+      complete_searched_facts = []
+
+      facts.each do |fact|
+        matched_facts = searched_facts.select { |searched_fact| fact.name.match(searched_fact.name) }
+        matched_fact = matched_facts.first
+        searched_fact = SearchedFact.new(fact.name,
```
sebastian-miclea: Maybe SearchedFact can receive a matched_fact object that also contains the fact name. In this case it will receive only one parameter for initialisation.

lib/facter.rb
```diff
@@ -20,51 +20,78 @@ def self.value(*args)
...
       threads
     end
 
-    def join_threads!(threads, searched_facts)
+    def create_fact(searched_fact)
+      fact_class = searched_fact.fact_class
+      if searched_fact.name.end_with?('.*')
+        fact_without_wildcard = searched_fact.name[0..-3]
+        filter_criteria = searched_fact.user_query.split(fact_without_wildcard).last
+        fact_class.new(searched_fact.filter_tokens).call_the_resolver(filter_criteria)
```
sebastian-miclea: I can see here a good point for having a base class for the Fact, this could contain the filter_tokens, and filter_criteria and then in the Fact implementation you will have them available without needing an `initialize` method.

sebastian-miclea (approved):

PR #28, opened 2019-09-02
(FACT-2006) Add rake task for generating fact templates.

BogdanIrimie:


tasks/fact_generator/fact_creator.rb
```diff
@@ -0,0 +1,69 @@
...
+      fact_name = fact['fact_name']
+
+      path = create_directory_path(operating_system, fact_name)
+      create_directory_structure(path)
+      create_fact_file(path, fact_name, operating_system)
+    end
+  end
+
+  def create_fact(operating_system, fact_name)
+    puts fact_name
```
ciprianbadescu: are we allowed to use **puts**?
I would expect to use logging instead.

tasks/fact_generator/fact_creator.rb
```diff
@@ -0,0 +1,69 @@
...
+class FactCreator
+  def create_facts
+    facts_json_string = File.read(File.join(ROOT_DIR, 'tasks', 'fact_generator', 'facts.json'))
+    facts = JSON.parse(facts_json_string)
+
+    facts.each do |fact|
+      operating_system = fact['os']
+      fact_name = fact['fact_name']
+
+      path = create_directory_path(operating_system, fact_name)
```
ciprianbadescu: should here be called create_fact instead duplicate code?

tasks/fact_generator/fact_creator.rb
IrimieBogdan: Replaced with logger.

tasks/create_fact.rake
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+desc 'Create a fact template for the specified arguments'
+task :create_fact, [:os, :fact_name] do |_, args|
```
gimmyxd: This should fail gracefully if os/fact_name are not passed

tasks/fact_generator/fact_creator.rb
IrimieBogdan: Replaced with call to create_fact

tasks/create_fact.rake
IrimieBogdan: Added graceful fail when no arguments are provided.

sebastian-miclea (approved):

PR #39, opened 2019-09-12
(FACT-2014) Added Os fact for AIX

sebastian-miclea:


lib/framework/utils/odm_query.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
```
oanatmaria: This class should be in resolvers/aix/aix_utils/ because is specific for AIX and it shouldn't be in framework

sebastian-miclea: I'm ok with having **utils** folder for each os that needs specific implementations, having **aix**/**aix**_utils is a bit redundant in my opinion.

BogdanIrimie (approved):

PR #40, opened 2019-09-13
(FACT-2020) Fix output for to_hash and value

BogdanIrimie:


lib/framework/core/facter.rb
```diff
@@ -2,29 +2,31 @@
...
 require 'pathname'
 
-ROOT_DIR = Pathname.new(File.expand_path('../..', __dir__)) unless defined?(ROOT_DIR)
+ROOT_DIR = Pathname.new(File.expand_path('../../..', __dir__)) unless defined?(ROOT_DIR)
 require "#{ROOT_DIR}/lib/framework/core/file_loader"
 
 module Facter
   def self.to_hash
-    Facter::Base.new.resolve_facts([])
+    resolved_facts = Facter::Base.new.resolve_facts({}, [])
```
ciprianbadescu: one alternative would be to have default value for resolve_facts arguments

IrimieBogdan: Added default values

ciprianbadescu: then here you can call without parameters and will use default value

PR #42, opened 2019-09-16
(FACT-2022)-Os name fact for Solaris

alexandru-d-popa:
Implementation for os name & kernel for solaris


lib/resolvers/solaris/os_resolver.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class OsResolver < BaseResolver
```
oanatmaria: tests for resolver?

loopinu: done.

BogdanIrimie (approved):

PR #48, opened 2019-09-20
(FACT-2028) Add Windows facts

oanatmaria:
Add domain, hostname and fqdn facts for Windows


lib/resolvers/windows/fqdn_resolver.rb
```diff
@@ -0,0 +1,57 @@
...
+          nil
+        end
+
+        def compose_fqdn(hostname, domain)
+          return [hostname, domain].join('.') if hostname && domain
+
+          hostname
+        end
+
+        def build_fact_list(fact_name)
```
IrimieBogdan: Should return a list of facts

lib/resolvers/windows/fqdn_resolver.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        def retrieve_domain
+          win = Win32Ole.new
+          result = win.exec_query('select DNSDomain from Win32_NetworkAdapterConfiguration where IPEnabled = True')
+          unless result
+            @log.debug 'WMI query returned no results for Win32_NetworkAdapterConfiguration with value DNSDomain.'
+            return
+          end
+          result.each do |network|
+            return network.DNSDomain if network.DNSDomain && !network.DNSDomain.empty?
```
IrimieBogdan: The condition could be replaced with
```
if network.DNSDomain.present?
```

lib/resolvers/windows/fqdn_resolver.rb
```diff
@@ -0,0 +1,59 @@
...
+            subscribe_to_manager
+            result || compose_fqdn(fact_name)
+          end
+        end
+
+        private
+
+        def retrieve_hostname
+          output, _status = Open3.capture2('hostname')
+          output ? output.strip : output
```
IrimieBogdan: This can be replaced with
```
output&.strip
```

lib/resolvers/windows/fqdn_resolver.rb
```diff
@@ -0,0 +1,59 @@
...
+          nil
+        end
+
+        def compose_fqdn(fact_name)
+          domain = retrieve_domain
+          hostname = retrieve_hostname
+          fqdn = nil
+          fqdn = [hostname, domain].join('.') if hostname && domain
+
+          result = { hostname: hostname, domain: domain, fqdn: fqdn || hostname }
```
sebastian-miclea: Shouldn't this be split in two resolvers hand have the logic that decides which (syscall/ole query) should it use?

lib/resolvers/windows/fqdn_resolver.rb
oanatmaria: undefined method `present?' for nil:NilClass

IrimieBogdan: present? is only available in rails

BogdanIrimie (approved):

PR #59, opened 2019-09-30
(FACT-2043) Migrate custom facts logic from Facter 2.x

BogdanIrimie:


lib/custom_facts/core/suitable.rb
```diff
@@ -20,7 +20,7 @@ module Suitable
       # @return [void]
       #
       # @api public
-      def weight?(weight)
+      def set_weight(weight)
```
gimmyxd: Do you still need this if you have the `attr_writer :weight` ?

IrimieBogdan: Removed the `attr_writer` and renamed methods for setting and accessing `weight`.

sebastian-miclea (approved):

PR #66, opened 2019-10-04
(FACT-2056) Zone fact for Solaris

alexandru-d-popa:
Added zone facts for Solaris

### DIFF
--- a/lib/resolvers/solaris/zone_resolver.rb
+++ b/lib/resolvers/solaris/zone_resolver.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class SolarisZone < BaseResolver
+      @log = Facter::Log.new
+      @semaphore = Mutex.new
+      @fact_list ||= {}
+      class << self
+        def resolve(fact_name)
+          @semaphore.synchronize do
+            result ||= @fact_list[fact_name]
+            subscribe_to_manager
+            result || build_zone_fact(fact_name)
+          end
+        end
+
+        private
+
+        def build_zone_fact(fact_name)
+          zone_adm_output, status = Open3.capture2('/usr/sbin/zoneadm list -cp')
+          unless status.to_s.include?('exit 0')
+            @log.debug("Command #{command} returned status: #{status}")
+            return
+          end
+          if zone_adm_output.empty?
+            @log.debug("Command #{command} returned an empty result")
+            return
+          end
+          @fact_list[:zone] = create_zone_facts(zone_adm_output)
+
+          @fact_list[fact_name]
+        end
+
+        def create_zone_facts(zones_result)
+          zones_fact = []
+          zones_result.each_line do |zone_line|
+            id, name, status, path, uuid, brand, ip_type = zone_line.split(':')
+            zone_fact = {
+              brand: brand,
+              id: id,
+              ip_type: ip_type.chomp,
+              name: name,
+              uuid: uuid,
+              status: status,
+              path: path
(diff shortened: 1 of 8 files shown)


lib/resolvers/solaris/zone_resolver.rb
```diff
@@ -0,0 +1,54 @@
...
+            result || build_zone_fact(fact_name)
+          end
+        end
+
+        private
+
+        def build_zone_fact(fact_name)
+          zone_name_output = CommandUtil.exec_command('/bin/zonename', @log)
+          zone_adm_output = CommandUtil.exec_command('/usr/sbin/zoneadm list -cp', @log)
+          return if zone_adm_output.nil? || zone_name_output.nil?
```
oanatmaria: return unless zone_adm_output || zone_name_output

lib/resolvers/solaris/zone_resolver.rb
```diff
@@ -0,0 +1,54 @@
...
+            subscribe_to_manager
+            result || build_zone_fact(fact_name)
+          end
+        end
+
+        private
+
+        def build_zone_fact(fact_name)
+          zone_name_output = CommandUtil.exec_command('/bin/zonename', @log)
+          zone_adm_output = CommandUtil.exec_command('/usr/sbin/zoneadm list -cp', @log)
```
sebastian-miclea: I see that the two sys calls are not dependent on each other. 
We started on a direction to have only one sys call per resolver to keep it as simple as possible.

Maybe this could be split in two resolvers, then the CommandUtil class would also not be necessary.

lib/resolvers/solaris/zone_resolver.rb
```diff
@@ -0,0 +1,54 @@
...
+          return unless zone_adm_output || zone_name_output
+
+          zones_result = zone_adm_output.split("\n")
+          zones_fact = create_zone_facts(zones_result)
```
IrimieBogdan: You could send `zone_adm_output` to `create_zone_fact` method and inside the method use `zones_result.each_line`

lib/resolvers/solaris/zone_resolver.rb
```diff
@@ -0,0 +1,54 @@
...
+          @fact_list[:zone] = zones_fact
+          @fact_list[:current_zone_name] = zone_name_output.chomp
+          @fact_list[fact_name]
+        end
+
+        def create_zone_facts(zones_result)
+          zones_fact = []
+          zone_regex_pattern = '(\\d+|-):([^:]*):([^:]*):([^:]*):([^:]*):([^:]*):([^:]*)'
+          zones_result.each do |zone_line|
+            id, name, status, path, uuid, brand, ip_type = zone_line.match(zone_regex_pattern).captures
```
IrimieBogdan: Output from `/usr/sbin/zoneadm list -cp` is `0:global:running:/::solaris:shared:-:none:`

Regex should not be used in this case, it is slow and makes the code harder to read. A simple `split(':')` would suffice.  The expression would become 
```
id, name, status, path, uuid, brand, ip_type = zone_line.split(':')
```

lib/resolvers/solaris/zone_resolver.rb
loopinu: I added a second rezolver for the zone_name but the CommandUtil class was created for reuse of the validation for the output and status from Open3.capture2.
I still think this class might be useful every time a sys call is used.

BogdanIrimie (approved):

sebastian-miclea (approved):

Issue #62, opened 2019-10-01
Implement path fact for Ubuntu


IrimieBogdan:
## Use Case
Implement `path` fact for Ubuntu

## Describe the Solution You Would Like
Implement `path` fact similar to the one in [Facter C++](https://github.com/puppetlabs/facter/blob/master/lib/src/facts/resolvers/path_resolver.cc) or [Facter Ruby](https://github.com/puppetlabs/facter/blob/2.x/lib/facter/path.rb)

## Additional Context
The fact can be generated using a rake task. E.g. 
```
rake 'create_fact[ubuntu,path]'
```

Example command
```
./bin/facter path
```

## Example output
```
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
```

PR #67, opened 2019-10-04
(FACT-2057) Add ubuntu path fact

AlexAvlonitis:
#62

### DIFF
--- a/lib/facts/ubuntu/path.rb
+++ b/lib/facts/ubuntu/path.rb
@@ -6,7 +6,7 @@ module Facter
       FACT_NAME = 'path'
 
       def call_the_resolver
-        fact_value = Facter::Resolvers::Path.resolve(FACT_NAME)
+        fact_value = Facter::Resolvers::Path.resolve(:path)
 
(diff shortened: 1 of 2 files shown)


IrimieBogdan:
Great job @AlexAvlonitis !

lib/facts/ubuntu/path.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Facter
+  module Ubuntu
+    class Path
+      FACT_NAME = 'path'
+
+      def call_the_resolver
+        fact_value = Facter::Resolvers::Path.resolve(FACT_NAME)
```
IrimieBogdan: It would be great if you could use a symbol here, in some cases the name we sent to the resolver is different from the fact name.
```suggestion
        fact_value = Facter::Resolvers::Path.resolve(:path)
```

AlexAvlonitis: @IrimieBogdan changed here: 8b0acd3 
Thanks :)

sebastian-miclea (approved):

Issue #68, opened 2019-10-04
Facterversion fact for Ubuntu


IrimieBogdan:
## Use Case
Implement `facterversion` fact for Ubuntu

## Describe the Solution You Would Like
Implement `facterversion` fact similar to the one for [windows](https://github.com/puppetlabs/facter-ng/blob/master/lib/facts/windows/facterversion.rb)

## Additional Context
The fact can be generated using a rake task. E.g. 
```
rake 'create_fact[ubuntu,facterversion]'
```

## Example command
```
./bin/facter facterversion
```

## Example output
```
3.11.9
```

## Validate implementation locally
```
./check.sh
```

PR #69, opened 2019-10-06
Add facterversion ubuntu fact

AlexAvlonitis:
#68 

It displays the `\n` as well i don't know if you want that sanitised maybe from the facter resolver?

Screenshot from 2019-10-06 14-03-27

### DIFF
--- a/lib/resolvers/facterversion_resolver.rb
+++ b/lib/resolvers/facterversion_resolver.rb
@@ -19,7 +19,7 @@ module Facter
 
         def read_version_file
           version_file = ::File.join(ROOT_DIR, 'VERSION')
-          @fact_list[:facterversion] = ::File.read(version_file).gsub("\n", '')
+          @fact_list[:facterversion] = ::File.read(version_file).chomp('')
         end
       end
     end
--- a/lib/resolvers/facterversion_resolver.rb
+++ b/lib/resolvers/facterversion_resolver.rb
@@ -19,7 +19,7 @@ module Facter
 
         def read_version_file
           version_file = ::File.join(ROOT_DIR, 'VERSION')
-          @fact_list[:facterversion] = ::File.read(version_file)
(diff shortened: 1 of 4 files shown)


IrimieBogdan:
It would be great if we can remove the `\n`.  You are right, the resolver is the best place to do it.
```
@fact_list[:facterversion] = ::File.read(version_file)
```

lib/resolvers/facterversion_resolver.rb
```diff
@@ -19,7 +19,7 @@ def resolve(fact_name)
 
         def read_version_file
           version_file = ::File.join(ROOT_DIR, 'VERSION')
-          @fact_list[:facterversion] = ::File.read(version_file)
+          @fact_list[:facterversion] = ::File.read(version_file).gsub("\n", '')
```
GabrielNagy: It's faster to use `tr` or `chomp` to remove the newline since it's a single character.

Using this benchmark: https://github.com/JuanitoFatas/fast-ruby/blob/master/code/string/gsub-vs-tr.rb

```
[~/tmp]$ ruby gsub-vs-tr.rb     [ruby-2.6.3] 
Warming up --------------------------------------
         String#gsub    94.593k i/100ms
           String#tr   268.495k i/100ms
Calculating -------------------------------------
         String#gsub      1.179M (± 3.5%) i/s -      5.959M in   5.060223s
           String#tr      4.431M (± 1.1%) i/s -     22.285M in   5.030466s

Comparison:
           String#tr:  4430565.7 i/s
         String#gsub:  1179430.0 i/s - 3.76x  slower
```

AlexAvlonitis: Nice! thx for the review, updated here: cbf0f6c

BogdanIrimie (approved):

IrimieBogdan:
Great job @AlexAvlonitis and @GabrielNagy, looks great!

Issue #75, opened 2019-10-08
Implement kernelmajversion fact for Ubuntu


IrimieBogdan:
## Use Case
Implement `kernelmajversion` fact for Ubuntu

## Describe the Solution You Would Like
Implement `kernelmajversion` fact similar to the one for [solaris](https://github.com/puppetlabs/facter-ng/blob/master/lib/facts/solaris/kernelmajversion.rb)

## Additional Context
The facts can be generated using a rake task. E.g. 
```
rake 'create_fact[ubuntu,kernelmajversion]'
```

## Example command
```
./bin/facter kernelmajversion
```

## Example output
```
19.0
```

## Validate implementation locally
```
./check.sh
```

IrimieBogdan:
@AlexAvlonitis, as we are quite a small number of contributors to this respo, your contribution with Ubuntu facts makes a [difference](https://docs.google.com/spreadsheets/d/1nLFU9_TEcSismnFqCySLBHdWfE-hcojuO54LWblGlGU/edit?usp=sharing).  If you want I can provide some facts that are more challenging (they require writing a resolver) .

AlexAvlonitis:
happy to help @IrimieBogdan when i have spare time :)

PR #77, opened 2019-10-09
Add kernelmajversion ubuntu fact

AlexAvlonitis:
#75

uname -r returns the release but uname -v doesn't , that's why i used `:kernelrelease` instead of `:kerneversion` param

Screenshot from 2019-10-09 12-09-28

Screenshot from 2019-10-09 12-04-21


IrimieBogdan:
Nice job @AlexAvlonitis , I just checked and `:kernelrelease` is perfect for obtaining the kernelmajversion fact.

BogdanIrimie (approved):

PR #86, opened 2019-10-14
(FACT-2069) Add path fact.

gigicampean:


spec/facter/facts/fedora/path_spec.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+describe 'Fedora Path' do
+  context '#call_the_resolver' do
+    it 'returns a fact' do
+      expected_fact = double(Facter::ResolvedFact, name: 'path', value: 'value')
```
IrimieBogdan: Would be nice to replace the `value` string with an actual PATH example.

BogdanIrimie (approved):

PR #93, opened 2019-10-15
(FACT-2061) Refactor custom facts

BogdanIrimie:


lib/framework/core/fact/internal/internal_fact_manager.rb
```diff
@@ -11,6 +13,10 @@ def resolve_facts(searched_facts)
 
     private
 
+    def filter_internal_facts(searched_facts)
+      searched_facts.select { |searched_fact| searched_fact.type == :core || searched_fact.type == :legacy }
```
sebastian-miclea: Imho this looks easier to read:
`[:core, :legacy].include? searched_fact.type`

sebastian-miclea (dismissed):

lib/framework/core/fact/internal/internal_fact_manager.rb
IrimieBogdan: Made the change here in `external_fact_manager` as  well.

sebastian-miclea (approved):

PR #99, opened 2019-10-16
(FACT-2076) Implemented kernelversion fact.

gigicampean:


lib/facts/fedora/kernelversion.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Facter
+  module Fedora
+    class Kernelversion
+      FACT_NAME = 'kernelversion'
+
+      def call_the_resolver
+        fact_value = Resolvers::Uname.resolve(:kernelversion)
```
IrimieBogdan: Resolvers::Uname.resolve(:kernelversion) returns `#1 SMP Tue May 14 19:33:09 UTC 2019` on Fedora. The value should be `5.0.16`

BogdanIrimie (approved):

PR #102, opened 2019-10-16
Networking domain fact ubuntu issue #78

m8051:
Attempts to implement #78


lib/resolvers/networking_domain_resolver.rb
```diff
@@ -0,0 +1,33 @@
...
+        def resolve(fact_name)
+          @semaphore.synchronize do
+            result ||= @fact_list[fact_name]
+            subscribe_to_manager
+            result || read_networking_domain(fact_name)
+          end
+        end
+
+        def read_networking_domain(fact_name)
+          output, _status = Open3.capture2('cat /etc/resolv.conf')
```
IrimieBogdan: Would be greta if you could use `File.open('/etc/resolv.conf')` instead of `Open3.capture2`.  We try to minimise the reliance on external tools if we can achieve the same thing within Ruby.

m8051: makes a lot of sense, will make the changes to use the File class, thanks for reviewing
it @IrimieBogdan

lib/resolvers/networking_domain_resolver.rb
```diff
@@ -0,0 +1,33 @@
...
+          output.each_line do |line|
+            next unless line.match(/search\s+(\S+)/)
```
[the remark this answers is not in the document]
m8051: nice catch and my bad, I should have tested it properly in Ubuntu, will make the changes.

Thanks @IrimieBogdan for that

m8051:
@IrimieBogdan committed a few more changes ^^ thanks for your time

m8051:
@IrimieBogdan plz see last commit whenever you have the chance with the changes required

BogdanIrimie (approved):

Issue #104, opened 2019-10-16
Implement kernelversion fact for Ubuntu


IrimieBogdan:
## Use Case
Implement `kernelversion` fact for Ubuntu

## Describe the Solution You Would Like
Implement `kernelversion` fact similar to the one for [solaris](https://github.com/puppetlabs/facter-ng/blob/master/lib/facts/solaris/kernelversion.rb). The `uname_resolver` should be called with `:kernelrelease` and the output should be parsed. Another option is to parse `/proc/version` file.

## Additional Context
The facts can be generated using a rake task. E.g. 
```
rake 'create_fact[ubuntu,kernelversion]'
```

## Example command
```
./bin/facter kernelversion
```

## Example output
```
4.15.0
```

## Validate implementation locally
```
./check.sh
```

joelmheim:
I'm looking into this one. Hope to provide a PR if I can work it out.

sebastian-miclea:
Excellent job! 
Thank you again for your contribution!

PR #110, opened 2019-10-25
Added kernelversion fact for Ubuntu.

joelmheim:
Should now show proper kernel version number on Ubuntu. Uses the Uname resolver and parses out the version numbers from that string.

Added a few tests to make sure that the parsing is handling a few cases.


joelmheim:
Fixes issue #104 .

sebastian-miclea:
> Should now show proper kernel version number on Ubuntu. Uses the Uname resolver and parses out the version numbers from that string.
> 
> Added a few tests to make sure that the parsing is handling a few cases.

Thank you for your contribution!

lib/facts/ubuntu/kernelversion.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Facter
+  module Ubuntu
+    class Kernelversion
+      FACT_NAME = 'kernelversion'
+
+      def call_the_resolver
+        version_numbers = Resolvers::Uname.resolve(:kernelrelease).scan(/\d+/)
+        fact_value = version_numbers[0..2].join '.'
```
sebastian-miclea: Could you please add brackets to the join method?

joelmheim: Will do. Sorry for the late reply. Have been travelling.

joelmheim:
Added parens to join. Sorry about the delay.

sebastian-miclea (approved):

Issue #116, opened 2019-10-28
Implement memory.swap.available_bytes fact for Ubuntu


IrimieBogdan:
## Use Case
Implement `memory.swap.available_bytes` fact for Ubuntu

## Describe the Solution You Would Like
Implement `memory.swap.available_bytes` fact similar to the one for [fedora](https://github.com/puppetlabs/facter-ng/blob/master/lib/facts/fedora/memory/swap/available_bytes.rb).

## Additional Context
The facts can be generated using a rake task. E.g. 
```
rake 'create_fact[ubuntu,memory.swap.available_bytes]'
```

## Example command
```
./bin/facter memory.swap.available_bytes
```

## Example output
```
847511552
```

## Validate implementation locally
```
./check.sh
```

henriquepjv:
@IrimieBogdan also done this related issue :smile:

IrimieBogdan:
@henriquepjv great job on the prompt PR #121 . There are multiple memory facts, very similar to one another, if you want have a look over them.

PR #121, opened 2019-10-30
Add memory swap available bytes for Ubuntu

henriquepjv:
[Issue_116](https://github.com/puppetlabs/facter-ng/issues/116)

### Why
- We still don't have the fact for Ubuntu memory swap available bytes

### How
- Add the fact and the test coverage for this new implementation


BogdanIrimie (approved):

IrimieBogdan:
@henriquepjv you did an excellent job on this PR.

PR #126, opened 2019-10-30
(FACT-2088) Create processors resolver facts for fedora.

gigicampean:


BogdanIrimie (dismissed):

lib/resolvers/processors_resolver.rb
```diff
@@ -0,0 +1,59 @@
...
+          def count_processors(tokens)
+            @fact_list[:processors] += 1 if tokens.first.strip == 'processor'
+          end
+
+          def construct_models_list(tokens)
+            @fact_list[:models] << tokens.last.strip if tokens.first.strip == 'model name'
+          end
+
+          def count_physical_processors(tokens)
+            @fact_list[:physical_processors].unshift(tokens.last.strip.to_i) if tokens.first.strip == 'physical id'
```
IrimieBogdan: `unshift ` https://apidock.com/ruby/Array/unshift does not offer any advantage here, it only makes the code harder to understand.

BogdanIrimie (approved):

PR #137, opened 2019-11-05
(FACT-2101) Created filesystems resolver for fedora.

gigicampean:


lib/resolvers/filesystems_resolver.rb
```diff
@@ -0,0 +1,33 @@
...
+            @semaphore.synchronize do
+              result ||= @fact_list[fact_name]
+              subscribe_to_manager
+              result || read_filesystems(fact_name)
+            end
+          end
+
+          def read_filesystems(fact_name)
+            output = File.read('/proc/filesystems')
+            array1 = []
```
IrimieBogdan: Maybe we can give a suggestive name for this array e.g. `filesystems`

gigicampean: Done :)

BogdanIrimie (approved):

PR #143, opened 2019-11-07
(FACT-2107) Created fips_enabled resolver for fedora.

gigicampean:


lib/resolvers/fips_enabled_resolver.rb
```diff
@@ -0,0 +1,30 @@
...
+              result ||= @fact_list[fact_name]
+              subscribe_to_manager
+              result || read_fips_file(fact_name)
+            end
+          end
+
+          def read_fips_file(fact_name)
+            file_output = File.read('/proc/sys/crypto/fips_enabled')
+            @fact_list[:fips_enabled] = 'false'
+            @fact_list[:fips_enabled] = 'true' if file_output.to_i == 1
```
IrimieBogdan: I would replace this code snippet with 
` @fact_list[:fips_enabled] = file_output.to_i == 1 ? true : false`

sebastian-miclea: I would replace it with
`@fact_list[:fips_enabled] = file_output == '1'`

spec/facter/resolvers/fips_enabled_resolver_spec.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+describe 'FipsEnabledResolver' do
+  let(:fips_enabled) { 'false' }
+
+  before do
+    allow(File).to receive(:read)
+      .with('/proc/sys/crypto/fips_enabled')
+      .and_return(load_fixture('fips_enabled').read)
```
oanatmaria: there is no need for a fixture file in this case. I would make a variable which you can set/unset and make a test scenario when is not fips_enabled also

spec/facter/resolvers/fips_enabled_resolver_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+describe 'FipsEnabledResolver' do
+  describe '#resolve' do
+    context 'when fips is not enabled' do
+      let(:fips_enabled) { false }
```
sebastian-miclea: I think we could drop this var and just add the value to the expect statement.

sebastian-miclea (approved):

PR #144, opened 2019-11-08
(FACT-2105) Add vmware fact for Windows

oanatmaria:


lib/resolvers/cpuid_source_resolver.rb
```diff
@@ -0,0 +1,54 @@
...
+            @fact_list[:xen] = check_for_xen(result)
+          end
+          @fact_list[fact_name]
+        end
+
+        def check_for_xen(result)
+          xen_vendor = 'XenVMMXenVMM'
+          max_entries = result[0]
+
+          return result[1].eql?(xen_vendor) if max_entries < 4 || max_entries >= 0x10000
```
IrimieBogdan: It would be nice to use decimal or hexadecimal (not combined) numbers in the same if statement.

BogdanIrimie (approved):

PR #145, opened 2019-11-12
(FACT-2116) Implemented dmi facts for fedora.

gigicampean:


gigicampean:
The resolver is not in "master" so it won't work unless the resolver is merged.

BogdanIrimie (approved):

PR #146, opened 2019-11-12
(FACT-2117) Implemented filesystems facts for fedora.

gigicampean:


gigicampean:
The resolver is not in "master" so it won't work unless the resolver will be merged.

sebastian-miclea (approved):

PR #155, opened 2019-11-14
(FACT-2107-suggestions) FIPS

BogdanIrimie:


spec/facter/resolvers/fips_enabled_resolver_spec.rb
```diff
@@ -1,14 +1,27 @@
 # frozen_string_literal: true
 
 describe 'FipsEnabledResolver' do
-  context 'when fips is not enabled' do
-    let(:fips_enabled) { 'false' }
+  describe '#resolve' do
+    context 'when fips is not enabled' do
+      let(:fips_enabled) { false }
```
sebastian-miclea: I think we could drop this var and just add the value to the expect statement.

PR #177, opened 2019-12-02
(FACT-2140) Added Options class, refactored FactManager

sebastian-miclea:


lib/framework/core/fact_manager.rb
```diff
@@ -13,9 +12,8 @@ def initialize
     end
 
     def resolve_facts(options = {}, user_query = [])
```
oanatmaria: maybe cli_options would be a better name

lib/framework/core/options/default_options.rb
```diff
@@ -10,16 +10,20 @@ def augment_with_defaults!
...
 
     def cli_defaults
-      @options[:debug] = false if @options[:debug].nil?
-      @options[:trace] = false if @options[:trace].nil?
-      @options[:verbose] = false if @options[:verbose].nil?
-      @options[:log_level] = 'error' unless @options[:log_level]
+      @options[:debug] = false
+      @options[:trace] = false
+      @options[:verbose] = false
+      @options[:log_level] = 'warn'
```
oanatmaria: this should be symbol

IrimieBogdan: Replaced with symbol.

lib/framework/core/fact_manager.rb
IrimieBogdan: Changed to cli_options

oanatmaria (approved):

PR #180, opened 2019-12-03
(FACT-2150) Block ruby fact when --no-ruby flag is true

BogdanIrimie:


lib/framework/core/options/helper_options.rb
```diff
@@ -4,7 +4,11 @@ module Facter
   module HelperOptions
     def augment_with_query_options!(user_query)
       @options[:user_query] = true if user_query.any?
-      @options[:custom_facts] = false if @options[:ruby] == false
+
+      if @options[:ruby] == false
```
sebastian-miclea: Could we add a test for this?

IrimieBogdan: Added tests to check black list and custom facts options.

sebastian-miclea (approved):

PR #183, opened 2019-12-09
(FACT-2152) Validate user options

oanatmaria:


lib/framework/core/options/options_validator.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+      end
+
+      @invalid_pairs_rules.each do |key, value|
+        values = [value]
+        any_value = ''
+        values.flatten.each { |val| any_value = val if options.include?(val) }
+        if options.include?(key) && !any_value.empty?
+          write_error_and_exit("#{key} and #{any_value} options conflict: please specify only one.")
+        end
```
IrimieBogdan: Can be written as
```
        common_values =  [value].flatten & options
        if options.include?(key) && common_values.size > 0
          write_error_and_exit("#{key} and #{common_values.first} options conflict: please specify only one.")
        end
```

BogdanIrimie (approved):

PR #185, opened 2019-12-10
(FACT-2178) Fix networking integrations tests

oanatmaria:


lib/framework/parsers/query_parser.rb
```diff
@@ -4,89 +4,97 @@ module Facter
...
-    def self.construct_loaded_fact(query_tokens, query_token_range, loaded_fact)
-      filter_tokens = query_tokens - query_tokens[query_token_range]
-      user_query = @query_list.any? ? query_tokens.join('.') : ''
-      fact_name = loaded_fact.name.to_s
-      klass_name = loaded_fact.klass
-      type = loaded_fact.type
-      SearchedFact.new(fact_name, klass_name, filter_tokens, user_query, type)
+      def construct_filter_tokens(query_tokens, query_token_range)
+        (query_tokens - query_tokens[query_token_range]).map do |token|
+          token !~ /\D/ && !token.empty? ? token.to_i : token.to_sym
```
IrimieBogdan: Instead of a double negation (not not digit) we should use ` token =~ /^[0-9]+/`

BogdanIrimie (dismissed):

spec/facter/resolvers/windows/networking_resolver_spec.rb
```diff
@@ -176,5 +173,75 @@
...
+        allow(NetworkUtils).to receive(:ignored_ip_address).with('fe80::7ca0:ab22:703a:b329')
+        allow(unicast).to receive(:[]).with(:Next).and_return(ptr)
+        allow(IpAdapterUnicastAddressLH).to receive(:new).with(ptr).and_return(unicast)
+        allow(unicast).to receive(:to_ptr).and_return(FFI::Pointer::NULL)
+        # back to build_interface_info
+        allow(NetworkUtils).to receive(:find_mac_address).with(adapter).and_return('00:50:56:9A:F8:6B')
+        # back to iterate_list
+        allow(adapter).to receive(:[]).with(:Next).and_return(ptr)
+        allow(IpAdapterAddressesLh).to receive(:new).with(ptr).and_return(adapter)
+        allow(adapter).to receive(:to_ptr).and_return(FFI::Pointer::NULL)
```
gimmyxd: We should have a dummy class that mimics al this behaviour to avoid this stubbing.

spec/facter/resolvers/windows/networking_resolver_spec.rb
```diff
@@ -28,16 +29,12 @@
...
-        allow(FFI::MemoryPointer).to receive(:new)
-          .with(IpAdapterAddressesLh.size, NetworkingFFI::BUFFER_LENGTH)
-          .and_return(adapter_address)
-        allow(NetworkingFFI).to receive(:GetAdaptersAddresses)
+        allow(FFI::MemoryPointer).to receive(:new).exactly(4).time
+                                                  .with(IpAdapterAddressesLh.size, NetworkingFFI::BUFFER_LENGTH)
+                                                  .and_return(adapter_address)
+        allow(NetworkingFFI)
+          .to receive(:GetAdaptersAddresses)
+          .exactly(3).time
```
gimmyxd: nit: `.times`

lib/facts/windows/ipaddress6.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Facter
+  module Windows
+    class Ipaddress6
+      FACT_NAME = 'ipaddress6'
+      FACT_TYPE = :legacy
+
+      def initialize(*args)
+        @log = Log.new
+        @log.debug 'Dispatching to resolve: ' + args.inspect
```
gimmyxd: This should interpolate, as it is faster than concatenation

lib/facts/windows/network6.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Facter
+  module Windows
+    class Network6
+      FACT_NAME = 'network6'
+      FACT_TYPE = :legacy
+
+      def initialize(*args)
+        @log = Log.new
+        @log.debug 'Dispatching to resolve: ' + args.inspect
```
gimmyxd: This should interpolate, as it is faster than concatenation

spec/facter/query_parser_spec.rb
```diff
@@ -18,6 +18,21 @@
...
+      os_name_class = Class.const_get('Facter::Ubuntu::OsName')
+
+      loaded_fact_os_name = double(Facter::LoadedFact, name: 'os.release', klass: os_name_class, type: :core)
+      loaded_facts = [loaded_fact_os_name]
+
+      matched_facts = Facter::QueryParser.parse(query_list, loaded_facts)
+
+      expect(matched_facts.size).to eq(1)
+      expect(matched_facts.first.fact_class).to eq(os_name_class)
+      expect(matched_facts.first.filter_tokens).to eq([:arry, 1, :val2])
```
gimmyxd: Usually you want to try to have a single expect per test. Writing tests like this, if the first assertion fails, you won't check the next ones.

You can group this in a context and have the general setup before running each expect.

http://www.betterspecs.org/#single

spec/facter/resolvers/windows/networking_resolver_spec.rb
```diff
@@ -176,5 +138,52 @@
...
+        {
+          address: 'fe80::7ca0:ab22:703a:b329',
+          netmask: IPAddr.new('ffff:ff00:0000:0000:0000:0000:0000:0000/ffff:ff00:0000:0000:0000:0000:0000:0000'),
+          network: IPAddr.new('fe80:0000:0000:0000:0000:0000:0000:0000/ffff:ff00:0000:0000:0000:0000:0000:0000')
+        }
+      end
+      before do
+        allow(IpAdapterAddressesLh).to receive(:read_list).with(adapter_address).and_yield(adapter)
+        allow(ptr).to receive(:read_wide_string_without_length).and_return('10.122.0.2')
+        allow(ptr).to receive(:read_wide_string_without_length).and_return('Ethernet0')
```
gimmyxd: why is this not set in the double? 
```let(:ptr) { double(FFI::Pointer, read_wide_string_without_length: 'Ethernet0') }```

File: spec/facter/query_parser_spec.rb [hunk 8]
```diff
@@ -18,6 +18,19 @@
       expect(matched_facts.first.fact_class).to eq(os_name_class)
     end
 
+    it 'compose filter_tokens correctly' do
+      query_list = ['os.release.arry.1.val2']
+
+      os_name_class = Class.const_get('Facter::Ubuntu::OsName')
+
+      loaded_fact_os_name = double(Facter::LoadedFact, name: 'os.release', klass: os_name_class, type: :core)
+      loaded_facts = [loaded_fact_os_name]
```
gimmyxd: `loaded_facts` should contain multiple entries, to make sure that only the correct facts are selected.
What if there is no LoadedFact with name containing `os.release`, that doesn't seem to be tested

spec/facter/query_parser_spec.rb
```diff
@@ -18,6 +18,19 @@
       expect(matched_facts.first.fact_class).to eq(os_name_class)
     end
 
+    it 'compose filter_tokens correctly' do
+      query_list = ['os.release.arry.1.val2']
+
+      os_name_class = Class.const_get('Facter::Ubuntu::OsName')
```
gimmyxd: do you need this extra variable? you can use `Facter::Ubuntu::OsName` directly

spec/facter/query_parser_spec.rb
oanatmaria: In 'creates one core searched fact' test above is this scenario covered

gimmyxd (approved):

PR #187, opened 2019-12-13
(FACT-2164) Update logger to use options for log level

BogdanIrimie:


lib/framework/logging/legacy_logger.rb
```diff
@@ -11,6 +11,11 @@ def level=(value)
       @stdout_logger.level = @stderr_logger.level = value
     end
 
+    def formatter=(format)
+      @stderr_logger.formatter = format
+      @stderr_logger.formatter = format
```
sebastian-miclea: Shouldn't one of these be **stdout_looger**?

sebastian-miclea (approved):

PR #189, opened 2019-12-16
(FACT-2200) Implemented dmi fact for OSX.

gigicampean:


BogdanIrimie (changes requested):
None of the facts work. The DMI resolver only works for Linux.

lib/framework/core/file_loader.rb
```diff
@@ -44,7 +44,7 @@ def load_lib_dirs(*dirs)
 os_hierarchy = CurrentOs.instance.hierarchy
 os_hierarchy.each { |operating_system| load_lib_dirs('facts', operating_system.downcase, '**') }
 
-load_lib_dirs('resolvers', os.to_s, '**') if os.to_s =~ /win|aix|solaris/
+load_lib_dirs('resolvers', os.to_s, '**') if os.to_s =~ /win|aix|solaris|macosx/
```
IrimieBogdan: In order to fix the merge conflict, you should take the version that is on master for file_loader

BogdanIrimie (changes requested):

lib/framework/core/file_loader.rb
gigicampean: The conflict was resolved ! :)

BogdanIrimie (approved):

PR #203, opened 2019-12-17
(FACT-2219) augeas fact for AIX

ciprianbadescu:


lib/resolvers/os_level_resolver.rb
```diff
@@ -18,8 +18,9 @@ def resolve(fact_name)
         end
 
         def read_oslevel(fact_name)
-          output, _status = Open3.capture2('oslevel -s')
+          output, _status = Open3.capture2('/usr/bin/oslevel -s 2>/dev/null')
           @@fact_list[:build] = output
+          @@fact_list[:kernel] = 'AIX'
```
gimmyxd: should this be hardcoded to AIX?

ciprianbadescu: This is my understanding from C facter: https://github.com/puppetlabs/facter/blob/master/lib/src/facts/aix/kernel_resolver.cc#L20

spec/facter/resolvers/augeas_resolver_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+describe 'OsLevel' do
```
oanatmaria: describe augeas resolver

ciprianbadescu: solved

oanatmaria (approved):

PR #205, opened 2019-12-17
(FACT-2246) Create mountpoints resolver

GabrielNagy:
Added dependency to `sys-filesystem` which uses FFI to get filesystem information. Also fixed a bug where `BytesToHumanReadable` did not add 2 digits after the decimal point.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,7 +5,7 @@ Metrics/LineLength:
   Max: 120
 
 Metrics/MethodLength:
-  Max: 20
+  Max: 25
   Exclude:
     - 'lib/custom_facts/util/values.rb'
     - 'lib/custom_facts/util/confine.rb'
@@ -23,6 +23,7 @@ Naming/ClassAndModuleCamelCase:
     - 'spec/mocks/**/*'
 
 Metrics/AbcSize:
+  Max: 30
   Exclude:
     - 'spec/custom_facts/util/parser_spec.rb'
     - 'spec/custom_facts/core/logging_spec.rb'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,7 +5,7 @@ Metrics/LineLength:
   Max: 120
 
 Metrics/MethodLength:
-  Max: 25
+  Max: 20
   Exclude:
     - 'lib/custom_facts/util/values.rb'
     - 'lib/custom_facts/util/confine.rb'
@@ -23,7 +23,6 @@ Naming/ClassAndModuleCamelCase:
     - 'spec/mocks/**/*'
 
 Metrics/AbcSize:
-  Max: 30
   Exclude:
     - 'spec/custom_facts/util/parser_spec.rb'
     - 'spec/custom_facts/core/logging_spec.rb'
--- a/lib/resolvers/mountpoints_resolver.rb
+++ b/lib/resolvers/mountpoints_resolver.rb
@@ -6,7 +6,7 @@ module Facter
       class Mountpoints < BaseResolver
         @semaphore = Mutex.new
         @fact_list ||= {}
-        @log = Facter::Log.new
+        @log = Facter::Log.new(self)
         class << self
           def resolve(fact_name)
             @semaphore.synchronize do
@@ -45,7 +45,7 @@ module Facter
             device
           end
 
-          def read_mounts
+          def read_mounts # rubocop:disable Metrics/AbcSize
             require 'sys/filesystem'
             mounts = []
             Sys::Filesystem.mounts do |fs|
--- a/lib/resolvers/mountpoints_resolver.rb
+++ b/lib/resolvers/mountpoints_resolver.rb
@@ -54,8 +54,7 @@ module Facter
               path = fs.mount_point
               options = fs.options.split(',')
 
-              next if path =~ %r{^/(proc|sys)} && filesystem != 'tmpfs'
-              next if filesystem == 'autofs'
+              next if path =~ %r{^/(proc|sys)} && filesystem != 'tmpfs' || filesystem == 'autofs'
 
               stats = Sys::Filesystem.stat(path)
               size_bytes = stats.bytes_total
--- a/lib/resolvers/mountpoints_resolver.rb
+++ b/lib/resolvers/mountpoints_resolver.rb
@@ -28,10 +28,10 @@ module Facter
             match&.captures&.first
           end
 
-          def compute_capacity(used, available, total)
-            if used == available
+          def compute_capacity(used, total)
+            if used == total
               '100%'
-            elsif used.positive? && total.positive?
+            elsif used.positive?
               "#{format('%.2f', 100.0 * used.to_f / total.to_f)}%"
             else
               '0%'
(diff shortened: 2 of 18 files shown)


GabrielNagy:
```sh-session
> ./bin/facter-ng mountpoints > mountpoints.facter-ng
> /usr/bin/facter mountpoints > mountpoints.facter
> diff -u mountpoints.facter mountpoints.facter-ng
--- mountpoints.facter	2019-12-17 15:51:52.406898271 +0200
+++ mountpoints.facter-ng	2019-12-17 15:51:49.073564901 +0200
@@ -1,8 +1,8 @@
 {
   / => {
-    available => "86.80 GiB",
-    available_bytes => 93201674240,
-    capacity => "80.02%",
+    available => "64.67 GiB",
+    available_bytes => 69435973632,
+    capacity => "84.31%",
     device => "/dev/nvme0n1p2",
     filesystem => "ext4",
     options => [
```

The only difference seems to be the root-reserved thing from https://github.com/puppetlabs/facter/pull/1857
[mountpoints.facter.txt](https://github.com/puppetlabs/facter-ng/files/3973580/mountpoints.facter.txt)
[mountpoints.facter-ng.txt](https://github.com/puppetlabs/facter-ng/files/3973581/mountpoints.facter-ng.txt)

.rubocop.yml
```diff
@@ -23,6 +23,7 @@ Naming/ClassAndModuleCamelCase:
     - 'spec/mocks/**/*'
 
 Metrics/AbcSize:
+  Max: 30
```
GabrielNagy: i would disable this cop entirely since it's really subjective based on the case

lib/resolvers/mountpoints_resolver.rb
```diff
@@ -0,0 +1,77 @@
...
+          end
+
+          def read_mounts
+            require 'sys/filesystem'
+            mounts = []
+            Sys::Filesystem.mounts do |fs|
+              device = fs.name
+              filesystem = fs.mount_type
+              path = fs.mount_point
+              options = fs.options.split(',')
```
sebastian-miclea: Are the options used anywhere?

lib/resolvers/mountpoints_resolver.rb
```diff
@@ -0,0 +1,77 @@
...
+
+              next if path =~ %r{^/(proc|sys)} && filesystem != 'tmpfs'
+              next if filesystem == 'autofs'
+
+              # If the "root" device, lookup the actual device from the kernel options
+              # This is done because not all systems symlink /dev/root
+              device = root_device if device == '/dev/root'
```
sebastian-miclea: The code would look a bit cleaner if the root_device method receives **device** as an argument, and decide inside the method what the device is.

lib/resolvers/mountpoints_resolver.rb
```diff
@@ -0,0 +1,77 @@
...
+
+              stats = Sys::Filesystem.stat(path)
+              size_bytes = stats.bytes_total
```
sebastian-miclea: You could send directly stats.bytes_total to the convert method without assigning to a local var.

lib/resolvers/mountpoints_resolver.rb
```diff
@@ -0,0 +1,77 @@
...
+            require 'sys/filesystem'
+            mounts = []
+            Sys::Filesystem.mounts do |fs|
+              device = fs.name
+              filesystem = fs.mount_type
+              path = fs.mount_point
+              options = fs.options.split(',')
+
+              next if path =~ %r{^/(proc|sys)} && filesystem != 'tmpfs'
+              next if filesystem == 'autofs'
```
sebastian-miclea: I'm not sure what type is **fs** is but, couldn't we reject these before iterating?

lib/resolvers/mountpoints_resolver.rb
GabrielNagy: The thing is, we want to call the `root_device` method only if our device is `/dev/root`. We can't have any other valid argument. We also don't care about `device` itself in the method.

lib/resolvers/mountpoints_resolver.rb
GabrielNagy: Yeah, they get put in the hash at line 69. They're part of `MOUNT_KEYS`.

lib/resolvers/mountpoints_resolver.rb
GabrielNagy: Makes sense, but `size_bytes` is also part of the final fact, so I need it in its own variable.

lib/resolvers/mountpoints_resolver.rb
sebastian-miclea: Might not phrased properly, but I would see this as a method that handles all devices, not only root. Basically, move the if inside and call it always. 

`device = compute_device(fs.name)`
This would reduce complexity of the **read_mounts** method

lib/resolvers/mountpoints_resolver.rb
GabrielNagy: I don't think so, we always get all mounts from `Sys::Filesystem.mounts`. We can drop unwanted mounts but then we'd have to iterate twice; first to reject and then to parse.

GabrielNagy:
I pushed some tests and reverted the `.rubocop.yml` changes. There's only the `Metrics/AbcSize` cop now.

```
Inspecting 769 files
.......................................................................................................................................................................................................................................................................................................................................................................C.........................................................................................................................................................................................................................................................................................................................................................................................................................

Offenses:

lib/resolvers/mountpoints_resolver.rb:48:11: C: Metrics/AbcSize: Assignment Branch Condition size for read_mounts is too high. [28.74/15]
          def read_mounts ...
          ^^^^^^^^^^^^^^^

769 files inspected, 1 offense detected
```

GabrielNagy:
Also specs are not running anymore after rebasing:

```
An error occurred while loading spec_helper.
Failure/Error:
  def initialize(logged_class)
    determine_callers_name(logged_class)
    set_format_for_file_logger
    set_format_for_stdout
  end

ArgumentError:
  wrong number of arguments (given 0, expected 1)
# ./lib/framework/logging/logger.rb:10:in `initialize'
# ./lib/resolvers/mountpoints_resolver.rb:9:in `new'
# ./lib/resolvers/mountpoints_resolver.rb:9:in `<class:Mountpoints>'
# ./lib/resolvers/mountpoints_resolver.rb:6:in `<module:Linux>'
# ./lib/resolvers/mountpoints_resolver.rb:5:in `<module:Resolvers>'
# ./lib/resolvers/mountpoints_resolver.rb:4:in `<module:Facter>'
# ./lib/resolvers/mountpoints_resolver.rb:3:in `<top (required)>'
# ./spec/mocks/kernel_mock.rb:7:in `require'
# ./spec/mocks/kernel_mock.rb:7:in `require'
# ./lib/framework/core/file_loader.rb:15:in `glob'
# ./lib/framework/core/file_loader.rb:15:in `load_dir'
# ./lib/framework/core/file_loader.rb:19:in `load_lib_dirs'
# ./lib/framework/core/file_loader.rb:39:in `<top (required)>'
# ./spec/mocks/kernel_mock.rb:7:in `require'
# ./spec/mocks/kernel_mock.rb:7:in `require'
# ./lib/facter-ng.rb:7:in `<top (required)>'
# ./spec/mocks/kernel_mock.rb:7:in `require'
# ./spec/mocks/kernel_mock.rb:7:in `require'
# ./spec/spec_helper.rb:21:in `<top (required)>'
No examples found.
No examples found.


Finished in 0.00003 seconds (files took 0.37146 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

sebastian-miclea (approved):

PR #207, opened 2019-12-17
(FACT-2255) Fix filtering for hashed with string keys

BogdanIrimie:


lib/framework/core/fact_filter.rb
```diff
@@ -7,9 +7,19 @@ module Facter
...
+        fact.value = symbolize_all_keys(fact.value)
+        fact.value = fact.filter_tokens.any? ? fact.value.dig(*fact.filter_tokens.map(&:to_sym)) : fact.value
       end
     end
+
+    private
+
+    def symbolize_all_keys(hash)
+      symbolized_hash = {}
+      hash.each do |k, v|
```
GabrielNagy: We should check if we're getting a hash before iterating. Sometimes we're just getting a string value.

```ruby
From: /home/gabi/repo/facter-ng/lib/framework/core/fact_filter.rb @ line 21 Facter::FactFilter#symbolize_all_keys:      
    17: def symbolize_all_keys(hash)
    18:   symbolized_hash = {}
    19:   require 'pry-byebug'
    20:   binding.pry
 => 21:   hash.each do |k, v|
    22:     symbolized_hash[k.to_sym] = v.is_a?(Hash) ? symbolize_all_keys(v) : v
    23:   end
    24:   symbolized_hash
    25: end

[1] pry(#<Facter::FactFilter>)> hash
=> "512.00 GiB"
[2] pry(#<Facter::FactFilter>)> hash.each {|k,v|}
NoMethodError: undefined method `each' for "512.00 GiB":String
from (pry):2:in `symbolize_all_keys'
```

GabrielNagy:
Confirmed https://github.com/puppetlabs/facter-ng/pull/205 works with this:

```sh-session
> ./bin/facter-ng mountpoints./
{
  available => "226.69 GiB",
  available_bytes => 243407589376,
  capacity => "4.82%",
  device => "/dev/sdb",
  filesystem => "ext4",
  options => [
    "rw",
    "relatime",
    "discard",
    "errors=remount-ro",
    "data=ordered"
  ],
  size => "250.98 GiB",
  size_bytes => 269490393088,
  used => "11.48 GiB",
  used_bytes => 12322131968
}
```

lib/framework/core/fact_filter.rb
IrimieBogdan: Added check to only symbolize keys if `fact.value` is a hash.

GabrielNagy:
Looks good :+1:, can we have a test for this?

IrimieBogdan:
Sure, added 2 tests.

GabrielNagy (approved):

PR #208, opened 2019-12-18
(FACT-2206) Kernelmajversion fact for OSX

Dorin-Pleava:
This commit ads kernelmajversion fact for osx

### DIFF
--- a/lib/facts/macosx/kernelmajversion.rb
+++ b/lib/facts/macosx/kernelmajversion.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Facter
+  module Macosx
+    class Kernelmajversion
+      FACT_NAME = 'kernelmajversion'
+
+      def call_the_resolver
+        kernel_major_version = Resolvers::Uname.resolve(:kernelrelease).match(/[0-9]+\.[0-9]+/).to_s
+
+        ResolvedFact.new(FACT_NAME, kernel_major_version)
+      end
+    end
+  end
+end
--- a/spec/facter/facts/macosx/kernelmajversion_spec.rb
+++ b/spec/facter/facts/macosx/kernelmajversion_spec.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+describe 'Macosx Kernelmajversion' do
+  context '#call_the_resolver' do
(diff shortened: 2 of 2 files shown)


lib/facts/macosx/kernelmajversion.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Facter
+  module Macosx
+    class Kernelmajversion
+      FACT_NAME = 'kernelmajversion'
+
+      def call_the_resolver
+        kernel_major_version = Resolvers::Uname.resolve(:kernelrelease).match(/[0-9]+\.[0-9]+/).to_s
```
gimmyxd: We should test that the  `match` here converts the value correctly. As i see in the test, this is not covered.

spec/facter/facts/macosx/kernelmajversion_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    it 'corectly resolves the fact' do
+      expected_fact = double(Facter::ResolvedFact, name: 'kernelmajversion', value: fact_value)
+      allow(Facter::Resolvers::Uname).to receive(:resolve).with(:kernelrelease).and_return(resolver_result)
+
+      fact = Facter::Macosx::Kernelmajversion.new.call_the_resolver
+      expect(fact.name).to eq(expected_fact.name)
+      expect(fact.value).to eq(expected_fact.value)
+    end
```
IrimieBogdan: An alternative could be:
```
    let(:value) { '18.7' }
    let(:resolver_result) { '18.7.0' }
    let(:expected_resolved_fact) { double(Facter::ResolvedFact, name: 'kernelmajversion', value: value) }
    subject(:fact) { Facter::Macosx::Kernelmajversion.new }

    before do
      expect(Facter::Resolvers::Uname).to receive(:resolve).with(:kernelrelease).and_return(resolver_result)
      expect(Facter::ResolvedFact).to receive(:new).with('kernelmajversion', value).and_return(expected_resolved_fact)
    end

    it 'returns kernelmajversion fact' do
      expect(fact.call_the_resolver).to eq(expected_resolved_fact)
    end
```

BogdanIrimie (approved):

PR #209, opened 2019-12-18
(FACT-2223) hypervisors fact for AIX

ciprianbadescu:
```
[0] [AIX] root@aix61-4:~/facter-ng # ./bin/facter-ng hypervisors
{
  lpar => {
    partition_name => "aix61-4",
    partition_number => 13
  }
}
```

### DIFF
--- a/lib/facts/aix/hypervisors.rb
+++ b/lib/facts/aix/hypervisors.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Facter
+  module Aix
+    class Hypervisors
+      FACT_NAME = 'hypervisors'
+
+      def call_the_resolver
+        ResolvedFact.new(FACT_NAME, fact_value)
+      end
+
+      private
+
+      def fact_value
+        {}.tap do |hash|
+          hash['wpar'] = wpar_fact if wpar_fact
+          hash['lpar'] = lpar_fact if lpar_fact
+        end
+      end
+
+      def wpar_fact
+        wpar_key = Resolvers::Lpar.resolve(:wpar_key)
+        return unless wpar_key&.positive?
+
+        { key: wpar_key, configured_id: Resolvers::Lpar.resolve(:wpar_configured_id) }
+      end
+
+      def lpar_fact
+        lpar_partition_number = Resolvers::Lpar.resolve(:lpar_partition_number)
+        return unless lpar_partition_number&.positive?
+
+        { partition_number: lpar_partition_number, partition_name: Resolvers::Lpar.resolve(:lpar_partition_name) }
+      end
+    end
+  end
+end
--- a/lib/resolvers/lpar_resolver.rb
+++ b/lib/resolvers/lpar_resolver.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class Lpar < BaseResolver
+      @semaphore = Mutex.new
+      @fact_list ||= {}
+
+      class << self
+        def resolve(fact_name)
+          @semaphore.synchronize do
+            result ||= @fact_list[fact_name]
+            subscribe_to_manager
+            result || read_lpar(fact_name)
+          end
+        end
+
+        def read_lpar(fact_name)
+          lpar_cmd = '/usr/bin/lparstat -i'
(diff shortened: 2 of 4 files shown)


lib/resolvers/lpar_resolver.rb
```diff
@@ -0,0 +1,60 @@
...
+        def resolve(fact_name)
+          @semaphore.synchronize do
+            result ||= @fact_list[fact_name]
+            subscribe_to_manager
+            result || read_lpar(fact_name)
+          end
+        end
+
+        def read_lpar(fact_name)
+          supports_wpar = supports_wpar?
```
gimmyxd: You don't need this assignment

lib/resolvers/lpar_resolver.rb
```diff
@@ -0,0 +1,60 @@
...
+        def supports_wpar?
+          supports_wpar = false
+          output, _status = Open3.capture2('/usr/bin/oslevel 2>/dev/null')
+          if output =~ /^(\d)\.(\d)/
+            major = Regexp.last_match(1).to_i
+            minor = Regexp.last_match(2).to_i
+            supports_wpar = true if major > 6 || (major == 6 && minor.positive?)
+          end
+          supports_wpar
+        end
```
gimmyxd: We should memoize the result in case this method is used multiple times, to avoid future expensive calls
https://thoughtbot.com/blog/ruby-memoization-and-alternatives

lib/facts/aix/hypervisors.rb
```diff
@@ -0,0 +1,33 @@
...
+
+module Facter
+  module Aix
+    class Hypervisors
+      FACT_NAME = 'hypervisors'
+
+      def call_the_resolver
+        fact_value = {}
+        fact_value['wpar'] = wpar_fact if wpar_fact
+        fact_value['lpar'] = lpar_fact if lpar_fact
```
gimmyxd: you could extract this into a method, to avoid the local variable.
```ruby
def fact_value
  {}.tap do |hash|
    hash['wpar'] = wpar_fact if wpar_fact
    hash['lpar'] = lpar_fact if lpar_fact
  end
end
```

lib/resolvers/lpar_resolver.rb
```diff
@@ -0,0 +1,60 @@
...
+            result || read_lpar(fact_name)
+          end
+        end
+
+        def read_lpar(fact_name)
+          supports_wpar = supports_wpar?
+          lpar_cmd = '/usr/bin/lparstat -i'
+          lpar_cmd += ' -W' if supports_wpar
+          output, _status = Open3.capture2(lpar_cmd)
+          output.sub('WPAR Key', "\nWPAR Key").each_line do |line|
```
gimmyxd: you could avoid creating new variable using `sub!`, like
```ruby
(output.sub!('WPAR Key', "\nWPAR Key") || output).each_line
```

lib/resolvers/lpar_resolver.rb
```diff
@@ -0,0 +1,60 @@
...
+            supports_wpar = true if major > 6 || (major == 6 && minor.positive?)
+          end
+          supports_wpar
+        end
+
+        def parse_line(line)
+          items = line.split(':')
+          k = items[0].strip
+          v = items[1].strip
+          [k, v]
```
gimmyxd: I think this is equivalent to : ```line.split(":").map(&:strip)```

lib/resolvers/lpar_resolver.rb
ciprianbadescu: removed

lib/resolvers/lpar_resolver.rb
ciprianbadescu: done

lib/resolvers/lpar_resolver.rb
ciprianbadescu: done

lib/resolvers/lpar_resolver.rb
ciprianbadescu: string are frozen in the file, don't know if it worth the change

ciprianbadescu: strings are frozen in entire file, don't know if it worth the change

PR #210, opened 2019-12-18
(maint) oslevel resolver fixes

ciprianbadescu:
- use absolute path
- redirect stderr to null as may mangle output
- use @facts instead @@facts

### DIFF
--- a/lib/resolvers/aix/os_level_resolver.rb
+++ b/lib/resolvers/aix/os_level_resolver.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class OsLevel < BaseResolver
+      @semaphore = Mutex.new
+      @fact_list ||= {}
+
+      class << self
+        def resolve(fact_name)
+          @semaphore.synchronize do
+            result ||= @fact_list[fact_name]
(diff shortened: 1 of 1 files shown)


lib/resolvers/os_level_resolver.rb
```diff
@@ -3,25 +3,24 @@
...
-
         def resolve(fact_name)
-          @@semaphore.synchronize do
-            result ||= @@fact_list[fact_name]
+          @semaphore.synchronize do
+            result ||= @fact_list[fact_name]
             subscribe_to_manager
             result || read_oslevel(fact_name)
           end
         end
```
oanatmaria: can you please make this method private, also

ciprianbadescu: done

lib/resolvers/os_level_resolver.rb
```diff
@@ -3,26 +3,26 @@
 module Facter
   module Resolvers
     class OsLevel < BaseResolver
```
IrimieBogdan: We should move this resolver in AIX directory as it is specific to AIX.

oanatmaria (dismissed):

oanatmaria (approved):

PR #211, opened 2019-12-18
(FACT-2155) Add alias mechanism

oanatmaria:


lib/facts/macosx/os/architecture.rb
```diff
@@ -4,10 +4,11 @@ module Facter
   module Macosx
     class OsArchitecture
       FACT_NAME = 'os.architecture'
+      ALIASES = 'architecture'
 
       def call_the_resolver
         fact_value = Resolvers::Uname.resolve(:machine)
-        ResolvedFact.new(FACT_NAME, fact_value)
+        [ResolvedFact.new(FACT_NAME, fact_value), ResolvedFact.new(ALIASES, fact_value, :legacy)]
```
gimmyxd: This behaves differently now when `ALIAS` is set. It will return an array of objects instead of a single object.
Shouldn't we be consistent and return the same data structure everywhere?

oanatmaria: Currently facter supports both types of returns (array of resolved facts and just one resolved fact)

BogdanIrimie (approved):

lib/framework/core/fact_filter.rb
```diff
@@ -21,5 +22,12 @@ def symbolize_all_keys(hash)
       end
       symbolized_hash
     end
+
+    def filter_legacy_facts!(resolved_facts)
+      resolved_facts.reject! { |fact| fact.type =~ /unknown/ }
+      return if Options.get[:show_legacy]
+
+      resolved_facts.reject! { |fact| fact.type =~ /legacy/ } unless Options.get[:user_query]
```
gimmyxd: I would let the fact to respond if it is legacy or any other kind.

Implement `legacy?`,  `core?`  and `unknown` methods on `ReslovedFact`.
You can easily select just de facts that you want like: ```resolved_facts.select! (&:core?)}```
I am more fore selecting the things that you know, instead of rejecting a specific fact.

lib/models/resolved_fact.rb
```diff
@@ -2,11 +2,15 @@
...
-    attr_accessor :name, :value, :user_query, :filter_tokens
+    attr_reader :name, :type
+    attr_accessor :user_query, :filter_tokens, :value
 
-    def initialize(name, value = '')
+    def initialize(name, value = '', type = :core)
       @name = name
       @value = value
+      logger = Log.new(self)
+      type =~ /core|legacy/ ? @type = type : logger.warn('The specified type is invalid!')
```
gimmyxd: I would move this logic in a private method. Also is it enough just to warn for an invalid type? Why not just raise error if the type is not known?

PR #213, opened 2019-12-18
(MAINT) Fix load_averages fact for Linux

luchihoratiu:


lib/resolvers/load_averages_resolver.rb
```diff
@@ -3,29 +3,31 @@
...
-            load_avrg = {}
-            load_avrg.store('1min', averages[0])
-            load_avrg.store('5min', averages[1])
-            load_avrg.store('15min', averages[2])
-            @fact_list[:loadavrg] = load_avrg
+
+            result = {}
+            result.store('1m', averages[0])
+            result.store('5m', averages[1])
+            result.store('15m', averages[2])
```
gimmyxd: You can create the whole hash, no need to add each value at a time.

```ruby
result = { 
  '1m'  => averages[0]
  '5m'  => averages[1]
  '15m' => averages[2]
}
```

lib/resolvers/load_averages_resolver.rb
```diff
@@ -3,29 +3,31 @@
...
+              result || read_load_averages_file(fact_name)
             end
           end
 
           private
 
-          def read_loadavrg_file(fact_name)
+          def read_load_averages_file(fact_name)
             output = File.read('/proc/loadavg')
             averages = output.split(' ')
```
gimmyxd: Split directly on `File.read('/proc/loadavg')`, no need to create useless variables

lib/resolvers/load_averages_resolver.rb
```diff
@@ -0,0 +1,36 @@
...
+  module Resolvers
+    module Linux
+      class LoadAverages < BaseResolver
+        @log = Facter::Log.new(self)
+        @semaphore = Mutex.new
+        @fact_list ||= {}
+        class << self
+          def resolve(fact_name)
+            @semaphore.synchronize do
+              result ||= @fact_list[fact_name]
```
oanatmaria: missing subscribe to manager

oanatmaria (approved):

PR #215, opened 2019-12-18
(FACT-2203) Identity fact for OSX

Dorin-Pleava:
This commit adds the identity fact to osx platforms

### DIFF
--- a/lib/framework/formatters/legacy_fact_formatter.rb
+++ b/lib/framework/formatters/legacy_fact_formatter.rb
@@ -8,6 +8,7 @@ module Facter
 
     def format(resolved_facts)
       user_queries = resolved_facts.uniq(&:user_query).map(&:user_query)
+      replace_nil_with_empty_string(resolved_facts)
 
       return if user_queries.count < 1
       return format_for_multiple_user_queries(user_queries, resolved_facts) if user_queries.count > 1
@@ -18,6 +19,10 @@ module Facter
     end
 
     private
+
+    def replace_nil_with_empty_string(resolved_facts)
+      resolved_facts.select { |fact| fact.value.nil? }.map! { |fact| fact.value = '' }
+    end
 
     def format_for_no_query(resolved_facts)
       @log.debug('Formatting for no user query')
@@ -47,7 +52,7 @@ module Facter
 
       return fact_value if fact_value.is_a?(String)
 
-      pretty_json = fact_value ? hash_to_facter_format(fact_value) : ''
+      pretty_json = hash_to_facter_format(fact_value)
 
       @log.debug('Remove quotes from value if it is a simple string')
       pretty_json.gsub(/^"(.*)\"/, '\1')
--- a/lib/resolvers/identity_resolver.rb
+++ b/lib/resolvers/identity_resolver.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class PosxIdentity < BaseResolver
+      @log = Facter::Log.new(self)
+      @semaphore = Mutex.new
+      @fact_list ||= {}
+
+      class << self
+        def resolve(fact_name)
+          @semaphore.synchronize do
+            result ||= @fact_list[fact_name]
+            subscribe_to_manager
+            result || retrieve_identity(fact_name)
+          end
+        end
+
+        private
+
+        def retrieve_identity(fact_name)
+          require 'etc'
+
+          login_info = Etc.getpwuid
+          @fact_list[:gid] = login_info.gid
+          @fact_list[:group] = Etc.getgrgid(login_info.gid).name
+          @fact_list[:privileged] = login_info.uid.zero?
+          @fact_list[:uid] = login_info.uid
+          @fact_list[:user] = login_info.name
+          @fact_list[fact_name]
+        end
+      end
+    end
+  end
+end
--- a/spec/facter/resolvers/identity_resolver_spec.rb
+++ b/spec/facter/resolvers/identity_resolver_spec.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'ostruct'
+
+describe 'IdentityResolver' do
+  before do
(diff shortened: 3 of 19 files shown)


lib/resolvers/identity_resolver.rb
```diff
@@ -0,0 +1,33 @@
...
+      class << self
+        def resolve(fact_name)
+          @semaphore.synchronize do
+            result ||= @fact_list[fact_name]
+            subscribe_to_manager
+            result || retrieve_identity(fact_name)
+          end
+        end
+
+        def retrieve_identity(fact_name)
```
oanatmaria: make this method private

lib/resolvers/identity_resolver.rb
```diff
@@ -0,0 +1,35 @@
...
+            result || retrieve_identity(fact_name)
+          end
+        end
+
+        private
+
+        def retrieve_identity(fact_name)
+          require 'etc'
+
+          passwd = Etc.getpwuid
```
GeorgeMrejea: I would like a better name, this objects contains more than password. Suggest `login_info`

GeorgeMrejea (changes requested):

lib/resolvers/identity_resolver.rb
```diff
@@ -0,0 +1,35 @@
...
+          @fact_list[:gid] = passwd.gid
+          @fact_list[:group] = Etc.getgrgid(passwd.gid).name
+          @fact_list[:privileged] = passwd.uid.zero?.to_s
+          @fact_list[:uid] = passwd.uid
+          @fact_list[:user] = passwd.name
+          @fact_list[fact_name]
```
oanatmaria: could you add some unit tests for this resolver?

lib/resolvers/identity_resolver.rb
Dorin-Pleava: Agreed. I used passwd as that is the name of the structure returned. https://docs.ruby-lang.org/en/2.0.0/Etc.html#method-c-getpwuid. Will change

lib/resolvers/identity_resolver.rb
```diff
@@ -0,0 +1,35 @@
...
+
+        private
+
+        def retrieve_identity(fact_name)
+          require 'etc'
+
+          passwd = Etc.getpwuid
+          @fact_list[:gid] = passwd.gid
+          @fact_list[:group] = Etc.getgrgid(passwd.gid).name
+          @fact_list[:privileged] = passwd.uid.zero?.to_s
```
GeorgeMrejea: privileged is boolean, remove to_s

Dorin-Pleava: Removing to_s will cause identity.privileged to not show up when executing `facter-ng identity.privileged`. It can be made to a boolean, but it will have to be converted in the fact to string.

GeorgeMrejea:
```
[0] [AIX] root@aix71-9:~ # /root/facter-ng/bin/facter-ng identity.privileged
true
[0] [AIX] root@aix71-9:~ # date
Tue Dec 24 05:00:13 PST 2019
[0] [AIX] root@aix71-9:~ #
```

GeorgeMrejea:
```
[0] [AIX] root@aix71-9:~ # /root/facter-ng/bin/facter-ng identity
{
  gid => 0,
  group => "system",
  privileged => true,
  uid => 0,
  user => "root"
}
```

Dorin-Pleava: I get the same result on a privileged system, but when identity.privileged returns false, nothing is displayed. That is why I convert it to string.
Can you try `facter-ng identity.privileged` on a system where it should return boolean false?

IrimieBogdan: Was fixed by a change in legacy formatter.

lib/framework/formatters/legacy_fact_formatter.rb
```diff
@@ -19,6 +20,10 @@ def format(resolved_facts)
 
     private
 
+    def replace_nil_with_empty_string(resolved_facts)
+      resolved_facts.select { |fact| fact.value.nil? }.map { |fact| fact.value = '' }
```
gimmyxd: You can do this in a single iteration: `resolved_facts.map { |fact| fact.nil? ? "" : fact }`
Do we want to return o copy of the facts or it is ok to update the values inplace?

BogdanIrimie (dismissed):

gimmyxd (dismissed):

GeorgeMrejea (dismissed):

spec/facter/resolvers/identity_resolver_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    result = Facter::Resolvers::PosxIdentity.resolve(:uid)
+
+    expect(result).to eq(501)
+  end
+
+  it 'returns identity USER' do
+    result = Facter::Resolvers::PosxIdentity.resolve(:user)
+
+    expect(result).to eq('test1.test2')
+  end
```
gimmyxd: nit: you can dry this a bit with shared_examples: https://relishapp.com/rspec/rspec-core/v/3-8/docs/example-groups/shared-examples

```ruby
shared_examples_for "a resolved fact" do |fact_name, value|
    subject { Facter::Resolvers::PosxIdentity.resolve(fact_name) }
    it { is_expected.to eql(value) }
  end

  describe 'GID' do
    it_behaves_like 'a resolved fact', :gid, 20
  end

  describe 'GROUP' do
    it_behaves_like 'a resolved fact', :group, 'staff'
  end
  ...
```

lib/framework/formatters/legacy_fact_formatter.rb
Dorin-Pleava: We want just the value of the fact to be `''`, not the whole fact object. Filtering first with select, and then changing the value seems to be easier.
Also I think it's OK to modify in place.

BogdanIrimie (approved):

PR #218, opened 2019-12-18
(Fact-2224) Identity fact for AIX

GeorgeMrejea:


IrimieBogdan:
@GeorgeMrejea  can I close this PR? My understanding is that we will go with #215 for both OS'es.

PR #221, opened 2019-12-19
(FACT-2192) Added --strict option

sebastian-miclea:


sebastian-miclea:
There still is a edge case that I need look into
`  % cfacter os abc --strict                                                                                                                                                     

2019-12-19 16:27:50.554516 ERROR puppetlabs.facter - fact "abc" does not exist.
abc =>
os => {...`

lib/framework/core/fact_manager.rb
```diff
@@ -17,6 +17,7 @@ def resolve_facts(options = {}, user_query = [])
 
       loaded_facts = @fact_loader.load(Options.get)
       searched_facts = QueryParser.parse(user_query, loaded_facts)
+      puts searched_facts.inspect
```
IrimieBogdan: Should be removed.

BogdanIrimie (dismissed):

BogdanIrimie (approved):

PR #223, opened 2019-12-23
(FACT-2213) processors fact for OSX

GeorgeMrejea:


lib/facts/macosx/processor_resolver.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class Processors < BaseResolver
```
oanatmaria: this should be in a Linux module

lib/facts/macosx/processor_resolver.rb
```diff
@@ -0,0 +1,66 @@
...
+      @log = Facter::Log.new(self)
+      @semaphore = Mutex.new
+      @fact_list ||= {}
```
oanatmaria: could you please write some tests for this resolver as well?

lib/facts/macosx/processor_resolver.rb
```diff
@@ -0,0 +1,66 @@
...
+          @fact_list[:models] = Array.new(@fact_list[:logicalcount].to_i).fill(brand)
+        end
+
+        def build_speed(value)
+          @fact_list[:speed] = convert_hz(value.split(': ').fetch(1).to_i)
+        end
+
+        def convert_hz(speed)
+          prefix = { 3 => 'k', 6 => 'M', 9 => 'G', 12 => 'T' }
+          power = Math.log10(speed).floor
```
oanatmaria: Math.log10(speed) this could be -Infinity if speed is 0, i think there should be a check for zero value

lib/facts/macosx/processor_resolver.rb
```diff
@@ -0,0 +1,66 @@
...
+      @fact_list ||= {}
+      @items = { logical_count: 'hw.logicalcpu_max',
+                 physical_count: 'hw.physicalcpu_max',
+                 brand: 'machdep.cpu.brand_string',
+                 speed: 'hw.cpufrequency_max' }
+      class << self
+        # Count
+        # Models
+        # PhysicalCount
+        # Speed
```
gimmyxd: Is this comment adding any value? I saw it in different places on different resolves. At least we should be consistent if we want to do it like this.

lib/facts/macosx/processor_resolver.rb
```diff
@@ -0,0 +1,66 @@
...
+module Facter
+  module Resolvers
+    class Processors < BaseResolver
+      @log = Facter::Log.new(self)
+      @semaphore = Mutex.new
+      @fact_list ||= {}
+      @items = { logical_count: 'hw.logicalcpu_max',
+                 physical_count: 'hw.physicalcpu_max',
+                 brand: 'machdep.cpu.brand_string',
+                 speed: 'hw.cpufrequency_max' }
```
gimmyxd: This is better suited as a constant, than an instance variable.

lib/facts/macosx/processor_resolver.rb
```diff
@@ -0,0 +1,66 @@
...
+          @fact_list[:logicalcount] = count.split(': ').fetch(1).to_i
+        end
+
+        def build_physical_count(count)
+          @fact_list[:physicalcount] = count.split(': ').fetch(1).to_i
+        end
+
+        def build_models(model)
+          brand = model.split(': ').fetch(1)
+          @fact_list[:models] = Array.new(@fact_list[:logicalcount].to_i).fill(brand)
```
gimmyxd: `Array.new(@fact_list[:logicalcount].to_i, brand)`

lib/facts/macosx/processor_resolver.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        def build_fact_list(result)
+          build_logical_count(result[0])
+          build_physical_count(result[1])
+          build_models(result[2])
+          build_speed(result[3])
+        end
+
+        def build_logical_count(count)
+          @fact_list[:logicalcount] = count.split(': ').fetch(1).to_i
```
gimmyxd: You should be consistent on array values retrieve, you use direct access on lines 35..38 and then use you use `fetch`. I'm for accessing directly the value at index - `arr[i]`

lib/facts/macosx/processor_resolver.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        private
+
+        def read_processor_data(fact_name)
+          output, _status = Open3.capture2("sysctl #{@items.values.join(' ')}")
+          build_fact_list(output.split("\n"))
+          @fact_list[fact_name]
+        end
+
+        def build_fact_list(result)
```
gimmyxd: result is a bit ambiguous, maybe something like `processors_data` would be more suited for this

lib/facts/macosx/processor_resolver.rb
```diff
@@ -0,0 +1,69 @@
...
+            @fact_list[:physicalcount] = count.split(': ')[1].to_i
+          end
+
+          def build_models(model)
+            brand = model.split(': ').fetch(1)
+            @fact_list[:models] = Array.new(@fact_list[:logicalcount].to_i, brand)
+          end
+
+          def build_speed(value)
+            @fact_list[:speed] = convert_hz(value.split(': ')[1].to_i)
```
oanatmaria: is this used anywhere? i don't see any Resolvers::Macosx::Processors.resolve(:speed) in any fact

oanatmaria: Also if speed is 0 i think we should return nil for this fact as it is clear that we couldn't resolve it properly

GeorgeMrejea: lib/facts/macosx/speed.rb 
fact_value = Resolvers::Macosx::Processors.resolve(:speed)

oanatmaria: yep, you're right i didn't see it :D

BogdanIrimie (approved):

PR #224, opened 2019-12-24
(FACT-2215) system_profiler fact for OSX

Dorin-Pleava:
This commit adds system_profiler fact to OSX platforms.

### DIFF
--- a/lib/resolvers/macosx/system_profiler_resolver.rb
+++ b/lib/resolvers/macosx/system_profiler_resolver.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class SystemProfiler < BaseResolver
+      @log = Facter::Log.new(self)
+      @semaphore = Mutex.new
+      @fact_list ||= {}
+      NAME_HASH = { boot_rom_version: 'Boot ROM Version',
+                    cores: 'Total Number of Cores',
+                    hardware_uuid: 'Hardware UUID',
+                    l2_cache_per_core: 'L2 Cache (per Core)',
+                    l3_cache: 'L3 Cache',
+                    memory: 'Memory',
+                    model_identifier: 'Model Identifier',
+                    model_name: 'Model Name',
+                    processor_name: 'Processor Name',
+                    processor_speed: 'Processor Speed',
+                    processors: 'Number of Processors',
+                    serial_number: 'Serial Number (system)',
+                    smc_version: 'SMC Version (system)',
+                    boot_mode: 'Boot Mode',
+                    boot_volume: 'Boot Volume',
+                    computer_name: 'Computer Name',
+                    kernel_version: 'Kernel Version',
+                    secure_virtual_memory: 'Secure Virtual Memory',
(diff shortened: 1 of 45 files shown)


Dorin-Pleava:
Travis fails because methods `retrieve_hardware_data` and `retrieve_software_data` have too many assignments. Can't think of a better or more elegant solution that does not imply a lot of alteration on the output to get the keys.

oanatmaria:
> Travis fails because methods `retrieve_hardware_data` and `retrieve_software_data` have too many assignments. Can't think of a better or more elegant solution that does not imply a lot of alteration on the output to get the keys.
You can try something like:
`module Facter
  module Resolvers
    class SystemProfiler < BaseResolver
      @log = Facter::Log.new(self)
      @semaphore = Mutex.new
      @fact_list ||= {}
      NAME_HASH = { boot_rom_version: 'Boot ROM Version',
                    cores: 'Total Number of Cores',
                    hardware_uuid: 'Hardware UUID',
                    l2_cache_per_core: 'L2 Cache (per Core)',
                    l3_cache: 'L3 Cache',
                    memory: 'Memory',
                    model_identifier: 'Model Identifier',
                    model_name: 'Model Name',
                    processor_name: 'Processor Name',
                    processor_speed: 'Processor Speed',
                    processors: 'Number of Processors',
                    serial_number: 'Serial Number (system)',
                    smc_version: 'SMC Version (system)',
                    boot_mode: 'Boot Mode',
                    boot_volume: 'Boot Volume',
                    computer_name: 'Computer Name',
                    kernel_version: 'Kernel Version',
                    secure_virtual_memory: 'Secure Virtual Memory',
                    system_version: 'System Version',
                    uptime: 'Time since boot',
                    username: 'User Name'}.freeze

      class << self
        def resolve(fact_name)
          @semaphore.synchronize do
            result ||= @fact_list[fact_name]
            subscribe_to_manager
            result || retrieve_system_profiler(fact_name)
          end
        end

        private

        def retrieve_system_profiler(fact_name)
          system_profiler = Hash[`system_profiler SPHardwareDataType SPSoftwareDataType`
                                       .scan(/.*:[ ].*$/)
                                       .collect!(&:strip)
                                       .map { |e| e.split(': ') }]
          NAME_HASH.each { |key, v| @fact_list[key] = system_profiler[v] }
          @fact_list[fact_name]
        end
      end
    end
  end
end
`

lib/resolvers/macosx/system_profiler_resolver.rb
```diff
@@ -0,0 +1,60 @@
...
+            result ||= @fact_list[fact_name]
+            subscribe_to_manager
+            result || retrieve_system_profiler(fact_name)
+          end
+        end
+
+        private
+
+        def retrieve_system_profiler(fact_name)
+          system_profiler = Hash[`system_profiler SPHardwareDataType SPSoftwareDataType`
```
oanatmaria: could you add a debug message before this with something like: "executing command: system_profiler SPSoftwareDataType SPHardwareDataType"

oanatmaria (approved):

PR #226, opened 2020-01-06
(FACT-2155) Resolve comments alias mechanism

oanatmaria:


gimmyxd (dismissed):
👍 
You could write some tests for the resolved_fact class and squash your commits

gimmyxd (approved):

PR #227, opened 2020-01-06
(FACT-2204) Is_virtual fact for OSX

Dorin-Pleava:
This commit adds the is_virtual fact for OSX platforms.


IrimieBogdan:
Resolved in #256

PR #231, opened 2020-01-07
(FACT-2229) load_averages for aix

ciprianbadescu:
depends on https://github.com/puppetlabs/puppet-runtime/pull/272

### DIFF
--- a/lib/facts/aix/load_averages.rb
+++ b/lib/facts/aix/load_averages.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module Facter
+  module Aix
+    class LoadAverages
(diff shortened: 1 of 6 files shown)


lib/resolvers/aix/ffi_helper.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
```
oanatmaria: can you please move this class in aix/utils and also not put in the resolvers module?

ciprianbadescu: done

oanatmaria (approved):

PR #234, opened 2020-01-08
(FACT-2261) Fix legacy fact formatter for nil values

BogdanIrimie:


luchihoratiu (approved):

lib/framework/formatters/legacy_fact_formatter.rb
```diff
@@ -47,6 +47,7 @@ def format_for_multiple_user_queries(user_queries, resolved_facts)
 
     def format_for_single_user_query(user_query, resolved_facts)
       @log.debug('Formatting for single user query')
+      replace_nil_with_empty_string(resolved_facts)
```
luchihoratiu: I manually tested the 'memory' fact on OSX and with this fix the output is (when swap is disabled):
```
{
  swap => {
    available => "",
    available_bytes => "",
    capacity => "",
    encrypted => "",
    total => "",
    total_bytes => "",
    used => "",
    used_bytes => ""
  },
  system => {
    available => "4.94 GiB",
    available_bytes => 5301116928,
    capacity => "84.57%",
    total => "32.00 GiB",
    total_bytes => 34359738368,
    used => "27.06 GiB",
    used_bytes => 29058621440
  }
}
```
Without this line it would be:
```
{
  system => {
    available => "5.01 GiB",
    available_bytes => 5374636032,
    capacity => "84.36%",
    total => "32.00 GiB",
    total_bytes => 34359738368,
    used => "26.99 GiB",
    used_bytes => 28985102336
  }
}
```
But then 'system.swap' would be back to:
```
null
```
We should further check this case.

luchihoratiu (changes requested):

luchihoratiu (dismissed):

luchihoratiu (approved):

PR #235, opened 2020-01-09
(FACT-2264) Add fips, scope6 and system32 fact

oanatmaria:


BogdanIrimie (approved):

lib/resolvers/windows/fips_resolver.rb
```diff
@@ -0,0 +1,36 @@
...
+              result || read_fact_from_registry(fact_name)
+            end
+          end
+
+          private
+
+          def read_fact_from_registry(fact_name)
+            reg = ::Win32::Registry::HKEY_LOCAL_MACHINE
+                  .open('System\\CurrentControlSet\\Control\\Lsa\\FipsAlgorithmPolicy')
+            reg.each { |name, _value| @fact_list[:fips_enabled] = reg[name] != 0 if name == 'Enabled' }
```
ciprianbadescu: could reg.each be replaced with reg.any or something similar to break loop at first key match (should be checked if faster before)

oanatmaria: I will solve this with another PR as this one is already merged

PR #245, opened 2020-01-14
(FACT-2253) Fix Fedora disks fact

oanatmaria:
- Fix Fedora disks fact
- Add additional check for dmi resolver


lib/resolvers/disk_resolver.rb
```diff
@@ -10,31 +10,33 @@ class Disk < BaseResolver
...
+            compose_file_paths
+
+            FILE_PATHS.each do |key, file|
+              @fact_list[:disks].each do |disk, value|
+                file_path = DIR + disk + file
+
+                next unless File.exist?(file_path)
+
+                result = File.read(file_path).strip
+                value[key] = file =~ /size/ ? result.to_i * 512 : result
```
GabrielNagy: Maybe it would be good to leave a comment regarding the block size here. It seems it should always be 512, but it's not really obvious: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/types.h?id=v4.4-rc6#n124

GabrielNagy (dismissed):

lib/resolvers/disk_resolver.rb
```diff
@@ -10,31 +10,35 @@ class Disk < BaseResolver
...
-            result = File.read(file_path).strip
-            @fact_list[fact_name] = fact_name.to_s =~ /size/ ? result.to_i * 1024 : result
             @fact_list[fact_name]
           end
 
-          def compose_file_path(fact_name)
-            block, file = fact_name.to_s.split('_')
-            DIR + block + FILE_PATHS[file.to_sym]
+          def compose_file_paths
+            output, _status = Open3.capture2("ls #{DIR}")
```
sebastian-miclea: **You could use Dir.entries** then you won't have to split by **\n**

lib/resolvers/disk_resolver.rb
```diff
@@ -10,31 +10,35 @@ class Disk < BaseResolver
...
 
           def post_resolve(fact_name)
             @fact_list.fetch(fact_name) { read_facts(fact_name) }
           end
 
           def read_facts(fact_name)
-            file_path = compose_file_path(fact_name)
-            return nil unless File.exist?(file_path)
+            @fact_list[:disks] = {}
+            compose_file_paths
```
sebastian-miclea: Could we rename this to something like build_disks_hash. 
From what I see, it has nothing to do with paths.

lib/resolvers/disk_resolver.rb
```diff
@@ -10,31 +10,35 @@ class Disk < BaseResolver
...
+
+            FILE_PATHS.each do |key, file|
+              @fact_list[:disks].each do |disk, value|
+                file_path = DIR + disk + file
```
sebastian-miclea: Please use File.join

sebastian-miclea (dismissed):

sebastian-miclea (approved):

PR #259, opened 2020-01-21
(maint) Decreased FFI version to be inline with runtime

sebastian-miclea:


facter-ng-with-dep/facter-ng-with-dep.gemspec
```diff
@@ -32,7 +32,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rubocop', '~> 0.74.0'
   spec.add_development_dependency 'rubycritic', '~> 4.1.0'
 
-  spec.add_runtime_dependency 'ffi', '~> 1.11'
+  spec.add_runtime_dependency 'ffi', '~> 1.9'
```
gimmyxd: This will install any ffi >= 1.9 and < 2.0 which will end-up installing 1.12.1 as the `~> 1.11` was already doing.

sebastian-miclea: True, but puppet-runtime contains 1.9.25 and it was out of range.

gimmyxd (approved):

PR #261, opened 2020-01-27
(FACT-2305) Add ipaddress fact for Aix

oanatmaria:


lib/framework/core/file_loader.rb
```diff
@@ -52,7 +52,8 @@ def load_lib_dirs(*dirs)
 os_hierarchy = OsDetector.instance.hierarchy
 os_hierarchy.each { |operating_system| load_lib_dirs('facts', operating_system.downcase, '**') }
 
-load_lib_dirs('resolvers', os.to_s, '**')
+load_lib_dirs('resolvers', os.to_s)
+load_lib_dirs('resolvers', os.to_s, '**') if os.to_s =~ /windows/ # this will be removed with another PR
```
IrimieBogdan: Maybe we could replace both lines with `load_lib_dirs('resolvers', os.to_s, **)`
Or even better with `os_hierarchy.each { |operating_system| load_lib_dirs('resolvers', operating_system.downcase, '**') }`

lib/resolvers/aix/networking_resolver.rb
```diff
@@ -0,0 +1,34 @@
...
+        class << self
+          private
+
+          def post_resolve(fact_name)
+            @fact_list.fetch(fact_name) { read_netstat(fact_name) }
+          end
+
+          def read_netstat(fact_name)
+            output, _status = Open3.capture2('netstat -rn')
+            output = output.split("\n").select! { |line| (line =~ /\s\s[0-9]+.[0-9]+.[0-9]+.[0-9]+|\s\s.*:[0-9a-f]+/) }
```
IrimieBogdan: `select!` could be replaced with `select`

BogdanIrimie (approved):

PR #263, opened 2020-01-28
(FACT-2304) Add ipaddress for solaris

florindragos:


lib/resolvers/solaris/ipaddress_resolver.rb
```diff
@@ -0,0 +1,49 @@
...
+              output, _status = Open3.capture2("ifconfig #{primary_interface}")
+              output.each_line do |str|
+                puts str
+                if str.strip =~ /inet\s(\d{1,3}.\d{1,3}.\d{1,3}.\d{1,3}) .+/
+                  ip = Regexp.last_match(1)
+                  break
+                end
+              end
+            end
+            @fact_list[fact_name] = ip
```
IrimieBogdan: Should be `@fact_list[:ip] = ip`. Otherwise, if somebody asked the resolver for a random fact e.g. `domain`, it will return the `ip` address instead of returning `nil`.

lib/resolvers/solaris/ipaddress_resolver.rb
```diff
@@ -0,0 +1,49 @@
...
+            @fact_list[fact_name]
+          end
+
+          def read_primary_interface
+            iface = nil
+            output, _status = Open3.capture2('route -n get default')
```
sebastian-miclea: Can't we grep for **interface**? Then we wouldn't have to loop through all the lines

BogdanIrimie (approved):

PR #266, opened 2020-01-29
(FACT-2274) Added debug api method

sebastian-miclea:


lib/framework/logging/legacy_logger.rb
```diff
@@ -21,7 +21,7 @@ def info(msg)
     end
 
     def debug(msg)
-      @stderr_logger.debug(msg)
+      @stdout_logger.debug(msg)
```
oanatmaria: debug should write on stderr

BogdanIrimie (approved):

PR #270, opened 2020-01-30
(FACT-2275) Add trace method API

oanatmaria:


lib/custom_facts/core/logging.rb
```diff
@@ -119,7 +119,7 @@ def format_exception(exception, message, trace)
 
         arr.concat(exception.backtrace) if trace
 
-        arr.flatten.join("\n")
+        "\e[31m#{arr.flatten.join("\n")}\e[0m"
```
sebastian-miclea: You cold add a constant for the red color, there is already one set for reset.

sebastian-miclea (dismissed):

BogdanIrimie (approved):

PR #271, opened 2020-01-31
(FACT-2310) Correctly detect centos

BogdanIrimie:


lib/framework/detector/os_detector.rb
```diff
@@ -49,15 +49,17 @@ def create_hierarchy(operating_system) # rubocop:disable Metrics/CyclomaticCompl
 
     case operating_system.to_sym
     when :ubuntu
-      %w[Debian Ubuntu]
+      %w[Debian]
     when :fedora
```
oanatmaria: for the moment wouldn't be better for conditions to be :fedora || :amzn || :rhel || :centos ?

oanatmaria (approved):

PR #275, opened 2020-02-03
(FACT-2325) Add debugging API method

oanatmaria:


lib/facter-ng.rb
```diff
@@ -64,7 +64,7 @@ def self.core_value(user_query)
   end
 
   def self.method_missing(name, *args, &block)
-    log = Facter::Log.new(self)
+    log = Log.new(self)
```
IrimieBogdan: Should be `@logger ||= Log.new(self)`

BogdanIrimie (approved):

PR #284, opened 2020-02-06
(FACT-2313) (FACT-2318) Extend Facter API with fact() and []

florindragos:


lib/facter-ng.rb
```diff
@@ -41,6 +45,13 @@ def self.debugging(debug_bool)
     Options.instance.change_log_level(debug_bool)
   end
 
+  def self.fact(name)
+    fact = LegacyFacter::Util::Fact.new(name)
+    val = value(name)
+    fact.add({}) { setcode { val } }
+    fact
```
oanatmaria: these methods should return an object of type Facter::Util::Fact not LegacyFacter....

oanatmaria (approved):

PR #292, opened 2020-02-12
(maint) Fix user query dig

oanatmaria:


lib/framework/utils/utils.rb
```diff
@@ -20,5 +20,22 @@ def self.split_user_query(user_query)
       queries = user_query.split('.')
       queries.map! { |query| query =~ /^[0-9]+$/ ? query.to_i : query }
     end
+
+    def self.deep_stringify_keys(hash)
+      {}.tap do |h|
+        hash.each { |key, value| key.class == Integer ? h[key] = map_value(value) : h[key.to_s] = map_value(value) }
```
sebastian-miclea: I think this would be easier to read:
`.....h["#{key}"] = map_value(value)`

gimmyxd: How i understand it `deep_stringify_keys` it must `stringify` all the keys, even Integers. Also some tests for all the `Utils` stuff would be nice.

lib/framework/utils/utils.rb
```diff
@@ -20,5 +20,22 @@ def self.split_user_query(user_query)
...
+      end
+    end
+
+    def self.map_value(collection)
```
gimmyxd: Is a bit confusing how `deep_stringify_keys` and `map_value` pass values from one to another and back. You could inspire from how rails implemented this:
```ruby
https://apidock.com/rails/Hash/_deep_transform_keys_in_object
def _deep_transform_keys_in_object(object, &block)
    case object
    when Hash
      object.each_with_object({}) do |(key, value), result|
        result[yield(key)] = _deep_transform_keys_in_object(value, &block)
      end
    when Array
      object.map { |e| _deep_transform_keys_in_object(e, &block) }
    else
      object
    end
  end
	
  https://apidock.com/rails/Hash/deep_transform_keys
  def deep_transform_keys(&block)
    _deep_transform_keys_in_object(self, &block)
  end

  https://apidock.com/rails/v4.0.2/Hash/deep_stringify_keys
  def deep_stringify_hash
    deep_transform_keys{ |key| key.to_s }
  end
```

oanatmaria: refactored code

BogdanIrimie (dismissed):

BogdanIrimie (approved):

PR #293, opened 2020-02-12
(FACT-2193) Fix trace.rb acceptance test

oanatmaria:


lib/custom_facts/core/logging.rb
```diff
@@ -119,7 +119,10 @@ def format_exception(exception, message, trace)
           arr << message
         end
 
-        arr.concat(exception.backtrace) if trace
+        if trace
+          arr << 'backtrace:'
```
sebastian-miclea: Could you add a test for this? It would be nice to catch this earlier and not when running acceptance tests.

oanatmaria: Done

lib/util/exceptions.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Facter
+  module Util
+    class ResolveCustomFactError < StandardError; end
```
gimmyxd: Error classes should not be Utils, you can either declare them in the Facter module or have them in an `erros/exceptions` folders

lib/custom_facts/core/resolvable.rb
```diff
@@ -76,7 +76,8 @@ def value
         nil
       rescue StandardError => e
         LegacyFacter.log_exception(e, "Error while resolving custom fact #{qualified_name}: #{e.message}")
-        exit 1
+        at_exit { exit 1 }
```
gimmyxd: do you need to `exit 1` if you're raising error? Is this error rescued somewhere else?

lib/custom_facts/util/fact.rb
```diff
@@ -163,7 +163,11 @@ def sort_by_weight(resolutions)
 
       def find_first_real_value(resolutions)
         resolutions.each do |resolve|
-          value = resolve.value
+          begin
+            value = resolve.value
+          rescue Facter::Util::ResolveCustomFactError => _e
```
gimmyxd: nit: you don't need `_e` if you're not going to use the error

oanatmaria: done

lib/util/exceptions.rb
oanatmaria: Done

BogdanIrimie (approved):

PR #307, opened 2020-02-24
(FACT-2168) Add debug log with blocked facts

oanatmaria:


.rubocop.yml
```diff
@@ -20,6 +20,7 @@ Metrics/ModuleLength:
   Max: 100
   Exclude:
     - 'lib/facter.rb'
+    - 'lib/facter-ng.rb'
```
gimmyxd: Instead of keep excluding files form rules we should consider refactoring the Facter module. For example at a first glance we could extract all the logging functionality in a separate module and use that in Facter.

lib/facter.rb
```diff
@@ -143,6 +146,10 @@ def self.version
     ::File.read(version_file).strip
   end
 
+  private_class_method def self.log_blocked_facts
```
gimmyxd: We could wrap the all the `self.` methods in a `self` class and use `private` to handle private methods.

lib/framework/core/fact_manager.rb
```diff
@@ -11,9 +11,7 @@ def initialize
       @log = Log.new(self)
     end
 
-    def resolve_facts(options = {}, user_query = [])
-      Log.level = options.get[:log_level]
-
+    def resolve_facts(_options = {}, user_query = [])
```
gimmyxd: `options` is not used

lib/framework/core/options.rb
```diff
@@ -60,9 +60,10 @@ def self.respond_to_missing?(name, include_private) end
 
     def initialize_options
       @options = {}
+      @options[:config] = @priority_options[:config]
```
gimmyxd: `@options = { config: @priority_options[:config] }`

lib/framework/core/options/validate_options.rb
```diff
@@ -4,9 +4,11 @@ module Facter
...
     def validate_configs
       conflicting_configs.each do |op|
-        if op.values[0] && op.values[1]
-          raise_error("#{op.keys[0]} and #{op.keys[1]} options conflict: please specify only one")
-        end
+        next unless op.values[0] && op.values[1]
+
+        message = "#{op.keys[0]} and #{op.keys[1]} options conflict: please specify only one"
+        OptionsValidator.write_error_and_exit(message) if @options[:is_cli]
+        raise_error(message)
```
gimmyxd: Is this tested anywhere?

lib/facter.rb
```diff
@@ -143,6 +146,10 @@ def self.version
...
+    block_list = BlockList.instance.block_list
+    @logger.debug("blocking collection of #{block_list.join("\s")} facts") if !block_list.empty? && Options[:block]
```
gimmyxd: you can use `.any?` instead of `!empty?`

lib/framework/core/fact_manager.rb
IrimieBogdan: Options argument should be removed.

.rubocop.yml
oanatmaria: This will be done under another ticket

IrimieBogdan: Created https://tickets.puppetlabs.com/browse/FACT-2421 for breaking Facter API in multiple modules

gimmyxd:
There are a lot of code/file changes by only 2 spec files updated. I don't thinks this is tested properly

BogdanIrimie (approved):

PR #310, opened 2020-02-25
(FACT-2388) Custom facts with dots will not be hierarchical

IrimieBogdan:


IrimieBogdan:
This was intended as a backup, in case someone had dot notation in their custom fact name.

PR #313, opened 2020-02-26
(FACT-2412) Add debian dmi facts

oanatmaria:


gimmyxd:
facts tests should be written following: https://github.com/puppetlabs/facter-ng/commit/18cff0a1fa0d5750fdcebfe839df254964000b9a

Also now that spec tests are name spaced based on module/class name there is an inconsistency between the structure in lib and the structure in spec

florindragos (approved):

.rubocop_todo.yml
```diff
@@ -20,7 +20,7 @@ RSpec/ContextWording:
 RSpec/ExpectInHook:
   Enabled: false
 
-# Offense count: 412
+# Offense count: 424
```
gimmyxd: This should go down or at least stay the same

IrimieBogdan: This is addressed in https://tickets.puppetlabs.com/browse/FACT-2418 and https://tickets.puppetlabs.com/browse/FACT-2419

BogdanIrimie (approved):

PR #315, opened 2020-02-26
(FACT-2410) Add debian disks fact

oanatmaria:


gimmyxd:
facts tests should be written following: https://github.com/puppetlabs/facter-ng/commit/18cff0a1fa0d5750fdcebfe839df254964000b9a

Also now that spec tests are name spaced based on module/class name there is an inconsistency between the structure in lib and the structure in spec

oanatmaria:
Tests are written according to the exemple.

Filipovici-Andrei (approved):

florindragos (approved):

PR #319, opened 2020-02-27
(FACT-2370) Fixed missing stub for tests

sebastian-miclea:


spec/facter/facter_spec.rb
```diff
@@ -69,6 +69,10 @@
     end
 
     context '--strict' do
+      before do
+        allow_any_instance_of(Facter::Options).to receive(:[]).with(:config)
```
gimmyxd: i think this should also work as `allow(Facter::Options.instance).to receive(:[]).with(:config)`

Filipovici-Andrei (approved):

PR #320, opened 2020-02-27
(FACT-2406) Fix custom and external facts loading.

BogdanIrimie:


lib/custom_facts/util/fact.rb
```diff
@@ -32,6 +32,7 @@ class Fact
       def initialize(name, options = {})
         @name = name.to_s.downcase.intern
 
+        @options = ::Facter::Utils.deep_copy(options)
```
gimmyxd: This should work evenif Facter is not globally namespaced: `Facter::Utils.deep_copy(@options.merge(options))`

lib/custom_facts/util/fact.rb
```diff
@@ -52,7 +53,7 @@ def initialize(name, options = {})
       #
       # @api private
       def add(options = {}, &block)
-        @options = options
+        @options = ::Facter::Utils.deep_copy(@options.merge(options))
```
gimmyxd: This should work evenif Facter is not globally namespaced: `Facter::Utils.deep_copy(@options.merge(options))`

lib/custom_facts/util/fact.rb
IrimieBogdan: Removed global namespace.

lib/custom_facts/util/fact.rb
IrimieBogdan: Removed global namespace.

gimmyxd (approved):

PR #325, opened 2020-02-28
(FACT-2418) Should load facts from nested modules

oanatmaria:


lib/framework/core/fact_loaders/class_discoverer.rb
```diff
@@ -9,13 +9,24 @@ def initialize
     end
 
     def discover_classes(operating_system)
-      os_module_name = Module.const_get("Facter::#{operating_system}")
+      os_module_name = Module.const_get("Facter::#{operating_system}") # TODO: change this to Facts
```
gimmyxd: TODO should also have a ticket

lib/framework/core/fact_loaders/class_discoverer.rb
```diff
@@ -9,13 +9,24 @@ def initialize
...
+      find_nested_classes(os_module_name, discovered_classes = [])
+      discovered_classes
     rescue NameError
       @log.error("There is no module named #{operating_system}")
       []
     end
+
+    def find_nested_classes(mod, discovered_classes)
+      mod.constants.each do |c|
+        if mod.const_get(c).is_a? Class
```
gimmyxd: 'c' is not descriptive, maybe something like `constant_name` is a better fit.

lib/framework/core/fact_loaders/class_discoverer.rb
```diff
@@ -9,13 +9,24 @@ def initialize
...
+          discovered_classes << mod.const_get(c)
+        elsif mod.const_get(c).is_a? Module
+          find_nested_classes(Module.const_get("#{mod.name}::#{c}"), discovered_classes)
```
gimmyxd: This might not work as desired per:
```
>> Class.is_a? Class   #=> true
>> Class.is_a? Module #=> true
>> Module.is_a? Module #=> true
>> Module.is_a? Class #=> true
```

```
>> Facter::Macosx.const_get(:DmiProductName).is_a? Class   #=> true
>> Facter::Macosx.const_get(:DmiProductName).is_a? Module #=> true
>> Facter::Macosx.const_get(:DmiProductName).is_a? Module #=> true
```

An alternative is:
```
>> Facter::Macosx.const_get(:DmiProductName).instance_of? Class   #=> true
>> Facter::Macosx.const_get(:DmiProductName).instance_of? Module #=> false
```

lib/framework/core/fact_loaders/class_discoverer.rb
gimmyxd: `is_a? Class` should be replaced with `instance_of? Class` to be more reliable

lib/framework/core/fact_loaders/class_discoverer.rb
oanatmaria: TODO will be resolved in FACT-2419

BogdanIrimie (approved):

lib/framework/core/fact_loaders/class_discoverer.rb
gimmyxd: I was a bit unclear here, i was expecting the `TODO` to also contain the ticket like: `TODO - FACT-2419: change this to Facts`

lib/framework/core/fact_loaders/class_discoverer.rb
```diff
@@ -9,13 +9,24 @@ def initialize
...
     rescue NameError
       @log.error("There is no module named #{operating_system}")
       []
     end
+
+    def find_nested_classes(mod, discovered_classes)
+      mod.constants.each do |constant_name|
+        if mod.const_get(constant_name).instance_of? Class
+          discovered_classes << mod.const_get(constant_name)
+        elsif mod.const_get(constant_name).instance_of? Module
```
gimmyxd: If dealing with modules it reads better like: `.is_a? Module`

PR #334, opened 2020-03-04
(FACT-2177) Use core value when custom fact value is nil

florindragos:


lib/custom_facts/util/collection.rb
```diff
@@ -136,19 +136,23 @@ def to_hash
...
-        core_value = Facter.core_value(name)
-        return core_value if weight <= 0 && core_value
-
-        value
+        core_value.nil? ? value : core_value
       end
 
       private
 
+      def extract_value_and_weight(fact)
```
sebastian-miclea: I don't value should be here at all.
Could we add the weight condition to the fact object?

IrimieBogdan: Removed the method and initialise used_resolution_weight in fact.

sebastian-miclea (approved):

PR #337, opened 2020-03-05
(FACT-2175) External facts write messages on stderr

oanatmaria:


lib/custom_facts/core/execution/base.rb
```diff
@@ -53,7 +53,8 @@ def execute(command, options = {})
 
             begin
               wait_for_child = true
-              out = `#{expanded_command}`.chomp
+              out, stderr, _status_ = Open3.capture3(expanded_command.to_s)
+              log_stderr_from_file(stderr, expanded_command)
```
gimmyxd: Should we always log to stderr or only if status is not success?

lib/custom_facts/core/execution/base.rb
```diff
@@ -78,9 +79,15 @@ def execute(command, options = {})
               end
             end
 
-            out
+            out.strip
           end
         end
+
+        def log_stderr_from_file(msg, command)
```
gimmyxd: this method should be private

lib/custom_facts/core/execution/base.rb
gimmyxd: i would put the conditional outside of this method and call this like
`log_stderr_from_file(stderr, expanded_command) unless stderr.nil? || stderr.empty?`

spec/custom_facts/core/execution/fact_manager_spec.rb
```diff
@@ -68,11 +68,26 @@ def handy_method
...
+        allow(Open3).to receive(:capture3).with('/bin/foo').and_return(['', 'some error'])
+        allow(Facter::Log).to receive(:new).with('foo').and_return(logger)
+        allow(subject).to receive(:expand_command).with('foo').and_return('/bin/foo')
+      end
+
+      it 'loggs warning messages on stderr' do
+        subject.execute('foo')
+        expect(logger).to have_received(:warn).with('some error')
+      end
+    end
```
gimmyxd:
```
spec/custom_facts/core/execution/fact_manager_spec.rb:82:9: C: RSpec/SubjectStub: Do not stub methods of the object under test.
        allow(subject).to receive(:expand_command).with('foo').and_return('/bin/foo')
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/custom_facts/core/execution/fact_manager_spec.rb:82:15: C: RSpec/NamedSubject: Name your test subject if you need to reference it explicitly.
        allow(subject).to receive(:expand_command).with('foo').and_return('/bin/foo')
              ^^^^^^^
spec/custom_facts/core/execution/fact_manager_spec.rb:86:9: C: RSpec/NamedSubject: Name your test subject if you need to reference it explicitly.
        subject.execute('foo')
        ^^^^^^^
```

lib/custom_facts/core/execution/base.rb
oanatmaria: If there are logs on stderr, facter should always logs them as warnings

lib/custom_facts/core/execution/base.rb
oanatmaria: done

lib/custom_facts/core/execution/base.rb
oanatmaria: put the conditions on the first line of log_stderr_from_file method in order to not create a new logger if there is nothing to log

.github/workflows/rtc.yml
```diff
@@ -0,0 +1,11 @@
...
+name: RuboCop TODO checker
+
+on: [pull_request]
+
+jobs:
+  rtc:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: gimmyxd/rtc-action@0.0.5
```
gimmyxd: This should be added in a separate PR

oanatmaria: Done

lib/custom_facts/core/execution/base.rb
```diff
@@ -53,7 +53,8 @@ def execute(command, options = {})
 
             begin
               wait_for_child = true
```
oanatmaria: this variable is out of scope as it doesn't have to wait for command to execute when we use Open3. This will be removed

sebastian-miclea (approved):

PR #338, opened 2020-03-05
(FACT-2431) Fix Legacy::Util::Fact name attribute

Filipovici-Andrei:
Changed the attribute accessor for the Fact name to attr_reader to match the behaviour in Facter 3.

### DIFF
--- a/spec/custom_facts/util/fact_spec.rb
+++ b/spec/custom_facts/util/fact_spec.rb
@@ -24,7 +24,6 @@ describe LegacyFacter::Util::Fact do
 
   describe '#name' do
     it 'changing the name raises error' do
-      fact.name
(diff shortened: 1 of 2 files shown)


spec/custom_facts/util/fact_spec.rb
```diff
@@ -22,6 +22,13 @@
     end
   end
 
+  describe '#name' do
+    it 'changing the name raises error' do
+      fact.name
```
oanatmaria: i don't think this method call is needed

Filipovici-Andrei: Correct. Fixed it.

sebastian-miclea (approved):

PR #342, opened 2020-03-06
(FACT-2435) Expose :expand as an option to execute command

florindragos:


lib/custom_facts/core/execution/base.rb
```diff
@@ -72,6 +67,24 @@ def log_stderr_from_file(msg, command)
           logger = Facter::Log.new(file_name)
           logger.warn(msg.strip)
         end
+
+        def builtin_command?(command)
+          `type #{command}`.chomp =~ /builtin/ ? true : false
```
gimmyxd: this can be shortened as 
```ruby
`type #{command}`.chomp.match?(/builtin/)
```

florindragos: that doesn't work on ruby 2.3

oanatmaria (approved):

PR #358, opened 2020-03-10
(FACT-2443) Add Solaris filesystems fact

oanatmaria:


lib/resolvers/solaris/filesystems_resolver.rb
```diff
@@ -0,0 +1,32 @@
...
+          def read_sysdef_file(fact_name)
+            return unless File.exist?('/usr/sbin/sysdef')
+
+            file_content, _status = Open3.capture2('/usr/sbin/sysdef')
+            files = file_content.split("\n").map do |line|
+              line.split('/').last if line =~ /^fs\.*/
+            end
+
+            @fact_list[:file_systems] = files.compact.sort.join(',')
+            @fact_list[fact_name]
```
Filipovici-Andrei: I think this line is not needed, because the method return is what you assigned to the @fact_list

lib/resolvers/solaris/filesystems_resolver.rb
```diff
@@ -0,0 +1,32 @@
...
+
+          def read_sysdef_file(fact_name)
+            return unless File.exist?('/usr/sbin/sysdef')
+
+            file_content, _status = Open3.capture2('/usr/sbin/sysdef')
+            files = file_content.split("\n").map do |line|
+              line.split('/').last if line =~ /^fs\.*/
+            end
+
+            @fact_list[:file_systems] = files.compact.sort.join(',')
```
Filipovici-Andrei: In which case do you get a nil value, to justify the compact call?

Filipovici-Andrei (changes requested):

lib/resolvers/solaris/filesystems_resolver.rb
oanatmaria: if i delete the last line you will get @fact_list[:file_systems] value even if you call the resolver with :anything_you_want for example

lib/resolvers/solaris/filesystems_resolver.rb
oanatmaria: compact is for empty string also, if you have something like fs/amd64/ if you split after '/' the last string will be an empty one

Filipovici-Andrei (approved):

PR #361, opened 2020-03-11
(FACT-2445) Add El legacy facts

oanatmaria:


Filipovici-Andrei (changes requested):

Filipovici-Andrei (approved):

lib/facts/el/processors/count.rb
```diff
@@ -5,10 +5,11 @@ module El
     module Processors
       class Count
         FACT_NAME = 'processors.count'
+        ALIASES = 'processorscount'
```
mwhahaha: The legacy var was processorcount not processorscount

PR #363, opened 2020-03-11
(FACT-2455) Add El Selinux facts

oanatmaria:


Filipovici-Andrei (changes requested):
Please also add the aliases for each fact

lib/resolvers/selinux_resolver.rb
```diff
@@ -14,23 +14,48 @@ class << self
...
+
+          @fact_list[:policy_version] = File.read(policy_file) if File.readable?(policy_file)
+          @fact_list[:enforced] = false
+          enforce = File.read(enforce_file) if File.readable?(enforce_file)
+          @fact_list[:current_mode] = if enforce.eql?('1')
+                                        @fact_list[:enforced] = true
+                                        'enforcing'
+                                      else
+                                        'permissive'
+                                      end
```
sebastian-miclea: I think having it written like this would be a bit more clear on what's happening.
`if enforce.eql?('1')
  @fact_list[:enforced] = true
  @fact_list[:current_mode] = 'enforcing'
else
  @fact_list[:enforced] = false
  @fact_list[:current_mode] = 'permissive'
end`

lib/resolvers/selinux_resolver.rb
```diff
@@ -14,23 +14,48 @@ class << self
...
+        end
+
+        def read_config_facts
```
sebastian-miclea: Could you rename this to _read_selinux_config_? It doesn't ready any facts.

lib/resolvers/selinux_resolver.rb
```diff
@@ -14,23 +14,48 @@ class << self
...
+        def read_mounts_file(fact_name)
           output, _s = Open3.capture2('cat /proc/self/mounts')
           @fact_list[:enabled] = false
 
           output.each_line do |line|
             next unless line =~ /selinuxfs/
 
             @fact_list[:enabled] = true
-            @fact_list[:mountpoint] = line
+            @fact_list[:mountpoint] = line.split("\s")[1]
```
Filipovici-Andrei: Should you add this mountpoint in the fact list if it's not returned by any fact?

Filipovici-Andrei (approved):

PR #364, opened 2020-03-11
(FACT-2339) Fix El disks fact

oanatmaria:


lib/resolvers/disk_resolver.rb
```diff
@@ -17,29 +17,43 @@ def post_resolve(fact_name)
           end
 
           def read_facts(fact_name)
-            @fact_list[:disks] = {}
             build_disks_hash
 
             FILE_PATHS.each do |key, file|
               @fact_list[:disks].each do |disk, value|
+                next unless File.readable?(File.join(DIR, disk, 'device'))
```
sebastian-miclea: Why should device be readable if it's never used?

sebastian-miclea (approved):

PR #366, opened 2020-03-12
(FACT-2456) Add distro facts

oanatmaria:


Filipovici-Andrei (dismissed):

Filipovici-Andrei (approved):

lib/resolvers/lsb_release_resolver.rb
```diff
@@ -15,18 +16,28 @@ class << self
...
+        def retrieve_facts(fact_name)
+          lsb_release_installed? if @fact_list[:lsb_release_installed].nil?
+          read_lsb_release_file if @fact_list[:lsb_release_installed]
+          @fact_list[fact_name]
+        end
+
+        def lsb_release_installed?
+          output, _status = Open3.capture2('which lsb_release')
+          @fact_list[:lsb_release_installed] = !output.empty?
+        end
```
gimmyxd: instead of storing into fact list i would cache `lsb_release_installed?` method
```ruby
def lsb_release_installed?
  @lsb_release_installed ||= begin
    output, _status = Open3.capture2('which lsb_release')
    @lsb_release_installed = !output.empty?
  end
end
```

PR #383, opened 2020-03-17
(FACT-2476) Differentiate between core facts and custom facts with nil value

BogdanIrimie:


lib/models/resolved_fact.rb
```diff
@@ -8,7 +8,9 @@ class ResolvedFact
     def initialize(name, value = '', type = :core)
       @name = name
       @value = Utils.deep_stringify_keys(value)
-      type =~ /core|legacy/ ? @type = type : (raise ArgumentError, 'The type provided for fact is not legacy or core!')
+
+      @type = type if type =~ /core|legacy|custom/
+      raise ArgumentError, 'The type provided for fact is not legacy or core!' unless type =~ /core|legacy|custom/
```
sebastian-miclea: We could remove one condition if we write it like this:
```
raise ArgumentError, 'The type provided for fact is not legacy or core!' unless type =~ /core|legacy|custom/
@type = type

File: lib/models/fact_collection.rb [hunk 2]
```diff
@@ -8,7 +8,11 @@ def initialize
 
     def build_fact_collection!(facts)
       facts.each do |fact|
-        bury(*fact.name.split('.') + fact.filter_tokens << fact.value)
+        if fact.type == :custom
```
sebastian-miclea: I think this should be the same as
```
next if fact.type == :core && fact.value.nil?
bury_fact(fact)

sebastian-miclea (approved):

PR #393, opened 2020-03-19
(FACT-2471) Add Linux partitions fact

oanatmaria:


oanatmaria:
Do not merge label is there because I need to write some unit tests for resolver.

spec/facter/resolvers/partitions_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+  end
+
+  context 'when /sys/block is readable' do
+    before do
+      allow(File).to receive(:readable?).with(sys_block_path).and_return(true)
+      allow(Dir).to receive(:entries).with(sys_block_path).and_return(sys_block_subdirs)
+    end
+
+    context 'when block have a device subdir' do
+      let(:sys_block_subdirs) { ['.', '..', 'sda'] }
```
Filipovici-Andrei: this let can be moved at the top of the test, and deleted from the other tests.

spec/facter/resolvers/partitions_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    it 'returns empty hash' do
+      result = Facter::Resolvers::Partitions.resolve(:partitions)
+
+      expect(result).to eq({})
+    end
+  end
+
+  context 'when /sys/block is readable' do
+    before do
```
Filipovici-Andrei: This before block should be deleted because it's being overwritten in the following contexts that contain before blocks with the same allow statements.

lib/resolvers/partitions.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        def post_resolve(fact_name)
+          @fact_list.fetch(fact_name) { read_partitions(fact_name) }
+        end
+
+        def read_partitions(fact_name)
+          @fact_list[:partitions] = {}
+          return {} unless File.readable?(BLOCK_PATH)
+
+          blocks = Dir.entries(BLOCK_PATH).reject { |dir| dir =~ /^\.+/ }
```
Filipovici-Andrei: Shouldn't this variable be named block_devices? Block doesn't quite suggest anything.

lib/resolvers/partitions.rb
```diff
@@ -0,0 +1,101 @@
...
+          blocks.each do |block|
+            block_path = "#{BLOCK_PATH}/#{block}"
+            if File.directory?("#{block_path}/device")
+              device_for_path_exists(block_path)
```
Filipovici-Andrei: device_for_path_exists, dm_for_path_exists and loop_for_path_exists don't check that something for path exists, they populate partitions, so I think they should be named, something like, extract_partitions_from_dm...path, ...loop

spec/facter/resolvers/partitions_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      end
+
+      it 'return partitions fact' do
+        result = Facter::Resolvers::Partitions.resolve(:partitions)
+
+        expect(result).to eq(parts)
+      end
+    end
+
+    context 'when block have a dm subdir' do
```
Filipovici-Andrei: it's has instead of have. The same for the other contexts.

spec/facter/resolvers/partitions_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      let(:sda_subdirs) do
+        ['/sys/block/sda/queue',
+         '/sys/block/sda/sda2',
+         '/sys/block/sda/sda2/stat',
+         '/sys/block/sda/sda2/dev',
+         '/sys/block/sda/sda2/uevent',
+         '/sys/block/sda/sda1']
+      end
+
+      let(:parts) do
```
Filipovici-Andrei: what is parts? Could you please give it a more suggestive name?

Filipovici-Andrei (changes requested):

Filipovici-Andrei (approved):

BogdanIrimie (approved):

PR #394, opened 2020-03-20
(FACT-2478) Fix reload custom facts mechanism

BogdanIrimie:
Differentiate between reading from fact collection and reloading the entire collection.

### DIFF
--- a/lib/custom_facts/util/collection.rb
+++ b/lib/custom_facts/util/collection.rb
@@ -11,6 +11,7 @@ module LegacyFacter
         @facts = {}
         @internal_loader = internal_loader
         @external_loader = external_loader
+        @loaded = false
       end
 
       # Return a fact object by name.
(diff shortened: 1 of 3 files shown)


lib/custom_facts/util/collection.rb
```diff
@@ -98,11 +98,25 @@ def external_facts
...
 
+      def invalidate_custom_facts
+        @valid_custom_facts = false
+      end
+
+      def reload_custom_facts
+        @loaded = false
+      end
+
+      @loaded = false
```
gimmyxd: I think this should be on initialize as a property of collection.

IrimieBogdan: Moved to initialize

gimmyxd:
As this is a fix for a bug that is breaking modules tests, there should be some tests for this functionality in facter.

gimmyxd (approved):

PR #396, opened 2020-03-20
(FACT-2425) Added unit tests for the BaseResolver module

Filipovici-Andrei:


spec/facter/resolvers/base_resolver_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      resolver.invalidate_cache
+
+      expect(resolver.resolve('fact2')).to eq('fact2' => 'value')
+    end
+  end
+
+  describe '#subscribe_to_manager' do
+    it 'calls the CacheManager subscribe method' do
+      allow(Facter::CacheManager).to receive(:subscribe).with(resolver)
```
oanatmaria: is this still needed?

oanatmaria: if yes can it be wrapped into a before statement?

Filipovici-Andrei: Yes, it can be wrapped into a before block.

BogdanIrimie (approved):

PR #398, opened 2020-03-23
(FACT-2231) Add AIX mountpoints fact

oanatmaria:


lib/facts/aix/mountpoints.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      def call_the_resolver
+        Facter::ResolvedFact.new(FACT_NAME, construct_mountpoints)
+      end
+
+      def construct_mountpoints
+        mountpoints = Facter::Resolvers::Aix::Mountpoints.resolve(:mountpoints)
+        sizes = Facter::Resolvers::Aix::FilesystemsSizes.resolve(:sizes)
+
+        mountpoints.each { |mount, _value| mountpoints[mount].merge!(sizes[mount]) if sizes[mount] }
```
Filipovici-Andrei: you can use value instead of mountpoint[mount]

lib/resolvers/aix/filesystems_sizes.rb
```diff
@@ -0,0 +1,47 @@
...
+        class << self
+          private
+
+          def post_resolve(fact_name)
+            @fact_list.fetch(fact_name) { execute_df_command(fact_name) }
+          end
+
+          def execute_df_command(fact_name)
+            @fact_list[:sizes] = {}
+            output, _status = Open3.capture2('df -P 2>/dev/null')
```
Filipovici-Andrei: shouldn't we process the status value? Same for mountpoints resolver

lib/resolvers/aix/filesystems_sizes.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    module Aix
+      class FilesystemsSizes < BaseResolver
```
Filipovici-Andrei: shouldn't system be with capital S

lib/resolvers/aix/filesystems_sizes.rb
```diff
@@ -0,0 +1,47 @@
...
+        @fact_list ||= {}
+
+        class << self
+          private
+
+          def post_resolve(fact_name)
+            @fact_list.fetch(fact_name) { execute_df_command(fact_name) }
+          end
+
+          def execute_df_command(fact_name)
```
Filipovici-Andrei: you are extracting the file system sizes, not just executing the df command. Probably you should have another 2 methods for executing the command and extracting the file systems.

lib/resolvers/aix/filesystems_sizes.rb
```diff
@@ -0,0 +1,47 @@
...
+            @fact_list[:sizes] = {}
+            output, _status = Open3.capture2('df -P 2>/dev/null')
+            output.split("\n").map do |line|
+              next if line =~ /Filesystem|-\s+-\s+-/
+
+              elem = line.split("\s")
```
Filipovici-Andrei: what does elem contain? probably a more suitable name is needed? Same applies for Mountpoints resolver

lib/resolvers/aix/filesystems_sizes.rb
```diff
@@ -0,0 +1,47 @@
...
+              compute_sizes(elem)
+              @fact_list[:sizes][elem.last].merge!(device: elem[0])
+            end
+
+            @fact_list[fact_name]
+          end
+
+          def compute_sizes(info)
+            available_bytes = info[3].to_i * 512
```
Filipovici-Andrei: maybe the 512 number could be moved into a meaningful named variable. Like this, it is like a magic number.

Filipovici-Andrei (changes requested):

Filipovici-Andrei (commented):
Could you have used something from FileSystemHelper from Resolvers Utils?

Filipovici-Andrei (approved):

PR #399, opened 2020-03-23
(FACT-2481) Fix caching

BogdanIrimie:


lib/resolvers/macosx/system_profiler_resolver.rb
```diff
@@ -30,14 +30,13 @@ class SystemProfiler < BaseResolver
 
       @log = Facter::Log.new(self)
       @semaphore = Mutex.new
+      @fact_list = {}
```
Filipovici-Andrei: In other resolvers you have @fact_list ||= {}

lib/resolvers/macosx/processor_resolver.rb
```diff
@@ -6,7 +6,7 @@ module Macosx
       class Processors < BaseResolver
         @log = Facter::Log.new(self)
         @semaphore = Mutex.new
-        @fact_list ||= {}
+        @fact_list = {}
```
Filipovici-Andrei: In other resolvers you have @fact_list ||= {}

Filipovici-Andrei (changes requested):
I think we are going in the wrong direction with the @fact_list and @semaphore. They shouldn't be tampered with in the resolvers. They should exist only in the BaseResolver. BaseResolver should be like a template, for what resolvers should do, and each resolver would implement just the post_resolve method. Caching, logging and synchronising should be dealt with in just ONE PLACE, in the BaseResolver.

lib/resolvers/macosx/processor_resolver.rb
IrimieBogdan: As `@fact_list` is initialise on class creation, there is no need for `||=`

lib/resolvers/macosx/system_profiler_resolver.rb
IrimieBogdan: As `@fact_list` is initialise on class creation, there is no need for `||=`

lib/resolvers/networking_linux_resolver.rb
```diff
@@ -4,12 +4,13 @@ module Facter
   module Resolvers
     class NetworkingLinux < BaseResolver
       @semaphore = Mutex.new
+      @fact_list ||= {}
```
gimmyxd: should this be `@fact_list = {}` as in other resolvers?

IrimieBogdan:
@Filipovici-Andrei `@semaphore` serialize access to one resolver, it should be declared and used at a resolver level, otherwise it would serialize access to all resolvers! Caching is done at the resolver level and should be kept at this level because serialisation is done at this level. The logger can be moved to base resolver.

lib/resolvers/networking_linux_resolver.rb
IrimieBogdan: Yes

Filipovici-Andrei (dismissed):

gimmyxd (approved):

PR #400, opened 2020-03-23
(FACT-2482) Fix processorcount legacy variable

mwhahaha:
PR #361 added the legacy fact names, however the processor count legacy
fact was incorrectly configured to be 'processorscount' when
'processorcount' is the name in the documentation.

https://puppet.com/docs/facter/3.7/core_facts.html#processorcount


oanatmaria (approved):

oanatmaria:
Hi, you are right, there was a typo. Thank you very much for your contribution!

PR #405, opened 2020-03-24
(maint) Fix trace acceptance test

oanatmaria:


BogdanIrimie (dismissed):

sebastian-miclea (dismissed):

.rubocop.yml
```diff
@@ -130,6 +130,10 @@ Naming/FileName:
     - 'lib/facter.rb'
     - 'agent/lib/facter-ng.rb'
 
+Style/TrivialAccessors:
+  Exclude:
+    - 'lib/facter.rb'
+
```
gimmyxd: instead of doing this, i suggest creating `attr_reader/writer` for the desired resources, it can also be done in `class << self`

lib/facter.rb
```diff
@@ -212,6 +214,7 @@ def trace?
     # @api public
     def trace(bool)
       LegacyFacter.trace(bool)
```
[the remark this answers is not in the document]
oanatmaria: LegacyFacter.log_exception is still used in other classes like collection.rb, so setting trace in LegacyFacter is still needed.

spec/facter/facter_spec.rb
```diff
@@ -345,4 +351,22 @@
       Facter.debugging?
     end
   end
+
+  describe '#log_exception' do
+    context 'when trace options is true' do
+      before do
+        Facter.instance_variable_set(:@trace, true)
```
gimmyxd: why not set it using `Facter.trace` method?

gimmyxd: also i think it should be set to false, after running the test

BogdanIrimie (approved):

PR #408, opened 2020-03-25
(FACT-2492) handle elementary OS as Debian

mihaibuzgau:


BogdanIrimie (dismissed):

lib/framework/detector/os_detector.rb
```diff
@@ -52,6 +52,8 @@ def create_hierarchy(operating_system)
     case operating_system.to_sym
     when :ubuntu
       %w[Debian]
+    when :elementary
+      %w[Debian]
```
gimmyxd: nit: are we going to do this for each distro that is based on other distros?
I think there are a lot of them that are based on ubuntu, maybe we can identify them better.

https://en.wikipedia.org/wiki/Category:Debian-based_distributions

mihaibuzgau: yes, working on a POC for this. Right now it would be great if we could unstuck facter for Elementary.

gimmyxd: 👍

BogdanIrimie (approved):

lib/framework/detector/os_detector.rb
igalic: how did the old versions of facter do this mapping?

oanatmaria: Hi, Facter 3 is using CMAKE variable CMAKE_SYSTEM_NAME (which is the result of uname -s) to differentiate at build time on which os is running. Elementary and Debian will be just 'Linux' in this case.  Facter 2 is using Facter::Core::Execution.exec("uname -s").

mlove-au: > yes, working on a POC for this. Right now it would be great if we could unstuck facter for Elementary.

@mihaibuzgau  Any ETA on this? I need to get facter working on Raspbian and could extend this yet again, but if you have a PR coming I could wait a bit for something cleaner.

IrimieBogdan: @mlove-au for the moment, please add Raspbian similar to ElementaryOS, the effort is low and you will be able to use Facter 4 with Raspbian. We are already working on the new mechanism, but it might take some time until we implement, test and release it.

PR #410, opened 2020-03-25
(FACT-2475) Fix os.release on Debian

oanatmaria:
- Read release fact from /etc/debian_version when os is Debian and from /etc/os-release otherwise instead of lsb_release
- Fix os.release on Debian 10


florindragos (dismissed):

lib/resolvers/os_release_resolver.rb
```diff
@@ -39,6 +39,7 @@ def read_os_release_file(fact_name)
 
           fill_fact_list(pairs)
           process_name
+          find_debian_release if @fact_list[:name] =~ /Debian/
```
IrimieBogdan: I would create a new resolver for Debian. For example OsRelease has the fallowing alternatives  RedHatRelease, SuseRelease. We should create something similar for Debian.

lib/facts/debian/os/distro/release.rb
```diff
@@ -9,20 +9,29 @@ class Release
...
+             Facter::ResolvedFact.new(ALIASES[1], release['major'], :legacy),
+             Facter::ResolvedFact.new(ALIASES[2], release['minor'], :legacy)]
+          end
+
+          private
+
+          def determine_release_for_os
+            os_name = Facter::Resolvers::OsRelease.resolve(:name)
+
+            if os_name =~ /Debian/
```
IrimieBogdan: Maybe we can use == instead of regex

BogdanIrimie (approved):

PR #418, opened 2020-03-27
(FACT-2503) Fix RSpec/NamedSubject offences

oanatmaria:


spec/custom_facts/core/execution/fact_manager_spec.rb
```diff
@@ -134,44 +136,48 @@ def handy_method
     end
 
     describe 'and the command is not present' do
+      subject(:executor) { Facter::Core::Execution::Posix.new }
```
sebastian-miclea: Why is this needed? It's already defined.

oanatmaria: done

sebastian-miclea (approved):

PR #422, opened 2020-03-31
(FACT-2477) Collect facts from alternative sources

oanatmaria:
Collect Debian os.distro facts from /etc/os-release instead of using lsb-release. Only lsb* legacy facts will be resolved using lsb-release.

### DIFF
--- a/lib/facts/debian/lsbdistcodename.rb
+++ b/lib/facts/debian/lsbdistcodename.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Facts
+  module Debian
+    class Lsbdistcodename
+      FACT_NAME = 'lsbdistcodename'
+      TYPE = :legacy
+
+      def call_the_resolver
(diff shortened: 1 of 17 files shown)


lib/facts/debian/lsbdistcodename.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Facts
+  module Debian
+    class Lsbdistcodename
+      FACT_NAME = 'lsbdistcodename'
```
IrimieBogdan, line 6: This is a legacy fact, but it will be detected as a core fact because it's name is specified in `FACT_NAME` and does not contain wildcards .

oanatmaria: add TYPE = :legacy for application to detect it correctly as a legacy fact

BogdanIrimie (approved):

PR #426, opened 2020-04-01
(FACT-2506) Generate changelog for release

oanatmaria:


facter.gemspec
```diff
@@ -37,6 +37,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'bundler', '~> 2.0'
   spec.add_development_dependency 'coveralls', '~> 0.8.23'
+  spec.add_development_dependency 'github_changelog_generator' if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('2.5.0')
```
gimmyxd, line 40: i would not have this here, i would create a `group: release` in `Gemfile` and add all release related gems there.

I don't want to install this when I do a bundle in CI

Rakefile
```diff
@@ -7,3 +7,35 @@ RSpec::Core::RakeTask.new(:spec)
...
+if Bundler.rubygems.find_name('github_changelog_generator').any?
+  require 'github_changelog_generator/task'
+
+  GitHubChangelogGenerator::RakeTask.new :changelog do |config|
+    raise "Set CHANGELOG_GITHUB_TOKEN environment variable eg 'export CHANGELOG_GITHUB_TOKEN=valid_token_here'" if Rake.application.top_level_tasks.include?('changelog') && ENV['CHANGELOG_GITHUB_TOKEN'].nil?
+
+    config.user = 'puppetlabs'
+    config.project = 'facter-ng'
+    config.since_tag = '4.0.13'
+    config.future_release = '4.0.14'
```
gimmyxd, line 20: this should be configurable, as rake task arguments.

Rakefile
```diff
@@ -7,3 +7,35 @@ RSpec::Core::RakeTask.new(:spec)
 Dir.glob(File.join('tasks/**/*.rake')).each { |file| load file }
 
 task default: :spec
+
+if Bundler.rubygems.find_name('github_changelog_generator').any?
```
gimmyxd, line 11: Would be nice to return something useful `like you don't have the gem installed, and what you need to do`

oanatmaria: Done

facter.gemspec
oanatmaria: Done

Rakefile
oanatmaria: since_tag will read VERSION file for the last tag

Gemfile
```diff
@@ -4,5 +4,9 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
 gemspec name: 'facter'
 
+group :development do
+  gem 'github_changelog_generator', require: false if Gem::Version.new(RUBY_VERSION.dup) >= Gem::Version.new('2.3.0')
+end
```
gimmyxd, line 9: My suggestion was to make this in a different group and optional, because most of the people that are doing development don't need `github_changelog_generator`. But i'm fine with this if the team decide so.

I was suggesting something like:
```ruby
group(:release, optional: true) do
...
end
```
Make it optional, means that by default will the gem not be installed, eg: CI
And then if i need to use it, i can use `bundle config set with 'release'`.

Rakefile
gimmyxd: I missed the part that the task is generated only if `github_changelog_generator`, i'm fine as it was before, without creating a new task assuming that release process is documented.

Also if you are deciding to make the `github_changelog_generator` then it is ok for the changelog rake task to not be shown.

Rakefile
```diff
@@ -7,3 +7,52 @@ RSpec::Core::RakeTask.new(:spec)
...
+  require 'github_changelog_generator/task'
+
+  GitHubChangelogGenerator::RakeTask.new  :changelog do |config|
+
+    raise "Set CHANGELOG_GITHUB_TOKEN environment variable eg 'export CHANGELOG_GITHUB_TOKEN=valid_token_here'" if Rake.application.top_level_tasks.include?('changelog') && ENV['CHANGELOG_GITHUB_TOKEN'].nil?
+    config.user = 'puppetlabs'
+    config.project = 'facter-ng'
+    config.since_tag = File.read('VERSION').strip
+    config.exclude_labels = ['maintenance']
+    config.header = "# Change log\n\nAll notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org)."
```
gimmyxd, line 22: this is defined also in line 26

.rubocop.yml
```diff
@@ -5,6 +5,16 @@ require:
   - rubocop-performance
   - rubocop-rspec
 
+inherit_mode:
+  merge:
+    - Exclude
```
gimmyxd, line 10: Why is this needed?

gimmyxd (commented):
Is this going to append to the existing changelog or regenarate everything each time.

.rubocop.yml
oanatmaria: Because i excluded Rakefile and Gemfile from AllCops, rubocop gets confused and will try to run against all gems, even the one that are dependancies. In order to prevent that from happening these lines are needed

Rakefile
oanatmaria: I think is better this way because it is always shown and if you don't have the github_changelog_generatorgem installed, the new task raise an error and tell you what to do

Rakefile
oanatmaria: Yes, in order to append without a tag, this line is now removed and the header will be an empty string.

Rakefile
oanatmaria: also future_release will be provided from keyboard

BogdanIrimie (dismissed):

gimmyxd (dismissed):

.rubocop.yml
```diff
@@ -5,6 +5,16 @@ require:
...
+
+AllCops:
+  TargetRubyVersion: '2.3'
+  Exclude:
+    - 'Rakefile'
+    - 'Gemfile'
+
```
IrimieBogdan, line 17: We should fix the rubocop errors in these files, not exclude the files

.rubocop.yml
sebastian-miclea: But, why did you remove Rakefile and Gemfile?

BogdanIrimie (approved):

PR #431, opened 2020-04-02
(FACT-2480) Implement Facter::FileUtil for safe read of files

Filipovici-Andrei:


lib/framework/core/file_loader.rb
```diff
@@ -37,6 +37,7 @@ def load_lib_dirs(*dirs)
 load_dir(['config'])
 
 load_lib_dirs('resolvers')
+load_lib_dirs('resolvers', 'utils')
```
oanatmaria, line 40: you can delete 
require "#{ROOT_DIR}/lib/resolvers/utils/fingerprint.rb"
require "#{ROOT_DIR}/lib/resolvers/utils/ssh.rb"
require "#{ROOT_DIR}/lib/resolvers/utils/filesystem_helper.rb"

from the lines above

Filipovici-Andrei: if I remove require "#{ROOT_DIR}/lib/resolvers/utils/filesystem_helper.rb" I get an error saying
uninitialized constant Facter::FilesystemHelper

lib/resolvers/utils/file_helper.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    module Utils
+      module FileHelper
```
IrimieBogdan, line 6: As this should be use in all instances when we need to read a file, I propose moving it to `/lib/util`

IrimieBogdan: `FileHelper` should be a class with static methods, modules are meant for sharing methods between classes.

lib/resolvers/partitions.rb
```diff
@@ -41,23 +42,27 @@ def extract_from_device(block_path)
...
           else
             populate_partitions("/dev/mapper/#{map_name}", block_path)
           end
         end
 
         def extract_from_loop(block_path)
-          populate_partitions("/dev/#{block_path}", block_path) if File.readable?("#{block_path}/loop/backing_file")
-          backing_file = File.read("#{block_path}/loop/backing_file").chomp
-          populate_partitions("/dev/#{block_path}", block_path, backing_file)
+          backing_file = Facter::Util::FileHelper.safe_read("#{block_path}/loop/backing_file").chomp
```
IrimieBogdan, line 54: As we are already in the `Facter` module, you can drop `Facter::` from all the places it is used, except tests.

lib/resolvers/solaris/filesystems_resolver.rb
```diff
@@ -15,11 +15,11 @@ def post_resolve(fact_name)
           end
 
           def read_sysdef_file(fact_name)
-            return unless File.readable?('/usr/sbin/sysdef')
+            file_content = Facter::Util::FileHelper.safe_readlines('/usr/sbin/sysdef', nil)
```
IrimieBogdan, line 18: This is an executable, reading it as a file will provide incorrect information. You should check if it executable `File.executable?` and than use `Open3.capture3` to execute it.

BogdanIrimie (approved):

PR #433, opened 2020-04-06
(FACT-2233) Add AIX partitons fact

oanatmaria:


lib/resolvers/aix/partitions.rb
```diff
@@ -0,0 +1,83 @@
...
+
+            lsl_content.flatten!.select! { |elem| elem }.map! { |elem| elem.delete("\s") }
+
+            Hash[*lsl_content]
+          end
+
+          def compute_size(info_hash)
+            physical_partitions = info_hash['PPs'].to_i
+            size_physical_partition = info_hash['PPSIZE']
+            exp = size_physical_partition[/mega/] ? 1024**2 : 1024**3
```
Filipovici-Andrei, line 76: I think these values should be defined in constants

oanatmaria: Done

lib/resolvers/aix/utils/odm_query.rb
```diff
@@ -27,9 +27,9 @@ def like(field, value)
     def execute
       result = nil
       REPOS.each do |repo|
-        break if result
+        break if result && !result.empty?
 
-        result, _s = Open3.capture2("#{query} #{repo}")
+        result, _stderr, _s = Open3.capture3("#{query} #{repo}")
```
sebastian-miclea, line 32: Why use capture3 if _stderr_ is not used?

oanatmaria: If stderr is not captured, it will be displayed when the command is executed

Filipovici-Andrei (approved):

PR #437, opened 2020-04-07
Fact 2536 Added Arch platform

Filipovici-Andrei:


GabrielNagy:
I was able to get Facter to run on Manjaro (which is an Arch-based distribution) by applying this patch:

```diff
diff --git lib/framework/detector/os_detector.rb lib/framework/detector/os_detector.rb
index a0c0860..40aac63 100644
--- lib/framework/detector/os_detector.rb
+++ lib/framework/detector/os_detector.rb
@@ -52,7 +52,7 @@ class OsDetector
     return [] unless operating_system
 
     case operating_system.to_sym
-    when :ubuntu
+    when :ubuntu, :manjaro
       %w[Debian]
     when :elementary
       %w[Debian]
```

I know we want to be specific on operating systems, but I don't know if this is a good enough reason to add another ~150 files to the project (most if not all of them being duplicated from Debian). Will the process be the same when adding other Linux-based operating systems, like Slackware, Gentoo, Alpine, etc... (which do not necessarily inherit from EL/Debian/SLES)? The codebase will quickly get clogged this way.

But then again, considering the current structure of the project I don't know how to simplify these OS additions in a somewhat logical way... :man_shrugging:

BogdanIrimie (approved):

lib/framework/detector/os_detector.rb
```diff
@@ -27,6 +27,8 @@ def detect
                     :solaris
                   when /aix/
                     :aix
+                  when /arch/
+                    :arch
```
GabrielNagy, line 31: Shouldn't this go under Linux, since it's a Linux distribution? Can you also add `manjaro` alongside `arch` so I would also have facts?

bastelfreak:
Hi,
I can confirm that this patch makes the previous message `ERROR Facter::ClassDiscoverer - There is no module named Arch` (mentioned in FACT-2536) disappear.

IrimieBogdan:
This approach to adding new OS is deprecated in favour of #470

PR #441, opened 2020-04-09
(FACT-2538) Don't save core and legacy facts in collection if they have no value

BogdanIrimie:


Filipovici-Andrei (changes requested):
Nice, at least one unit test would be better😆

IrimieBogdan:
Added unit tests.

lib/models/fact_collection.rb
```diff
@@ -8,7 +8,7 @@ def initialize
 
     def build_fact_collection!(facts)
       facts.each do |fact|
-        next if fact.type == :core && fact.value.nil?
+        next if (fact.type == :core || fact.type == :legacy) && fact.value.nil?
```
IrimieBogdan, line 11: Didn't use `fact.type =~ /core|legacy/` because regex are slow

gimmyxd: How about `%i[core, legacy].include?(fact.type)` ?

IrimieBogdan: Great suggestion, made the change!

spec/facter/model/fact_collection_spec.rb
```diff
@@ -1,29 +1,74 @@
...
+
+        expect(fact_collection).to eq(expected_hash)
+      end
+    end
+
+    context 'when fact value is nil' do
+      context 'when fact type is legacy' do
+        let(:resolved_fact) { Facter::ResolvedFact.new('os.version', nil, :legacy) }
+
+        before do
```
Filipovici-Andrei, line 28: this before block is already defined

spec/facter/model/fact_collection_spec.rb
```diff
@@ -1,29 +1,74 @@
...
+          expected_hash = {}
+
+          expect(fact_collection).to eq(expected_hash)
+        end
+      end
+
+      context 'when fact type is core' do
+        let(:resolved_fact) { Facter::ResolvedFact.new('os.version', nil, :core) }
+
+        before do
```
Filipovici-Andrei, line 44: this before block is already defined

spec/facter/model/fact_collection_spec.rb
```diff
@@ -1,29 +1,74 @@
...
+          resolved_fact.filter_tokens = []
+          resolved_fact.user_query = 'os'
+        end
+
+        it 'does not add fact to collection' do
```
Filipovici-Andrei, line 49: this test repeats, it can be added into a shared_example

spec/facter/model/fact_collection_spec.rb
```diff
@@ -1,29 +1,74 @@
...
+
+      context 'when fact type is :custom' do
+        let(:resolved_fact) { Facter::ResolvedFact.new('operatingsystem', nil, :custom) }
 
-  it 'does not add elements to fact collection if fact value is nil' do
-    fact_collection = Facter::FactCollection.new
-    resolved_fact = Facter::ResolvedFact.new('os.version', nil)
-    resolved_fact.filter_tokens = []
-    resolved_fact.user_query = 'os'
+        before do
```
Filipovici-Andrei, line 60: this before block is already defined and the user_query can be put in a let block

Filipovici-Andrei (changes requested):

spec/facter/model/fact_collection_spec.rb
IrimieBogdan: I would avoid `shared_example`  as it can become confusing very fast

spec/facter/model/fact_collection_spec.rb
IrimieBogdan: Moved before block in the upper context.

spec/facter/model/fact_collection_spec.rb
IrimieBogdan: Moved before block in the upper context

spec/facter/model/fact_collection_spec.rb
IrimieBogdan: Moved before block in the upper context

spec/facter/model/fact_collection_spec.rb
```diff
@@ -1,29 +1,70 @@
...
+        expect(fact_collection).to eq(expected_hash)
+      end
+    end
+
+    context 'when fact value is nil' do
+      context 'when fact type is legacy' do
+        let(:fact_name) { 'os.version' }
+        let(:fact_value) { nil }
+        let(:type) { :legacy }
+        # let(:resolved_fact) { Facter::ResolvedFact.new('os.version', nil, :legacy) }
```
oanatmaria, line 33: remove commented line

IrimieBogdan: Removed commented line

oanatmaria (approved):

Filipovici-Andrei (approved):

PR #445, opened 2020-04-13
(FACT-2539) Create OS hierarchy and make all linux distributions inherit from it

IrimieBogdan:


GabrielNagy:
:+1: i have FACTS!!

PR #449, opened 2020-04-14
(FACT-2530) Fix for tests/facts/os_processors_and_kernel.rb

oanatmaria:
Description of the problem: On Ubuntu os.release should be {full: 18.04, major: 18.04} instead of {full: 18.04, major: 18, minor: 4}(as. it is currently). On Sles release should be padded with 0: 15.0 instead of 15. On Aix and Solaris processors fact is missing.
Fix description: Fixed release on ubuntu and sles. Aix and Solaris facts will be added with a different PR.


lib/resolvers/os_release_resolver.rb
```diff
@@ -36,6 +36,7 @@ def read_os_release_file(fact_name)
 
           fill_fact_list(pairs)
           process_name
+          alter_version_id
```
sebastian-miclea, line 39: I would rename this to **pad_version_id**. Alter is very generic.

oanatmaria: done

BogdanIrimie (approved):

PR #451, opened 2020-04-14
(FACT-2552) Add Solaris processors facts

oanatmaria:
Description: Added solaris processors facts that were retrieved with kstat command. Bringing all methods that converts (bytes to mb, bytes to human readable, to hz) into one class.


lib/facts_utils/converter.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module Facter
+  module FactsUtils
+    class Converter
+      class << self
+        def to_mb(value_in_bytes)
+          (value_in_bytes / (1024.0 * 1024.0)).round(2)
+        rescue NoMethodError
```
sebastian-miclea, line 9: We should have the same guard as in the other two methods to be consistent.

lib/facts_utils/converter.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module Facter
+  module FactsUtils
+    class Converter
```
sebastian-miclea, line 5: Can we rename this to **UnitConverter** to be less generic?

lib/facts_utils/converter.rb
```diff
@@ -0,0 +1,56 @@
...
+      class << self
+        def to_mb(value_in_bytes)
+          (value_in_bytes / (1024.0 * 1024.0)).round(2)
+        rescue NoMethodError
+          nil
+        end
+
+        def to_hz(speed)
```
sebastian-miclea, line 13: To be consistent, this should be named **hertz_to_human**

lib/facts_utils/converter.rb
```diff
@@ -0,0 +1,56 @@
...
+          speed = speed.to_i
+          return if !speed || speed.zero?
+
+          prefix = { 3 => 'k', 6 => 'M', 9 => 'G', 12 => 'T' }
+          power = Math.log10(speed).floor
+          validated_speed = power.zero? ? speed.to_f : speed.fdiv(10**power)
+          format('%<displayed_speed>.2f', displayed_speed: validated_speed).to_s + ' ' + prefix[power] + 'Hz'
+        end
+
+        def bytes_to_human_readable(bytes)
```
sebastian-miclea, line 23: I would remove **_readable**, it's as clear even without it.

lib/facts_utils/converter.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        private
+
+        def pad_number(number)
+          number = number.to_s
+          number << '0' if number.split('.').last.length == 1
+          number
+        end
+
+        def determine_exponent(bytes)
```
sebastian-miclea, line 43: Could we move units/prefix in this method? That would clear up the **bytes_to_human_readable** method.

lib/facts_utils/converter.rb
```diff
@@ -0,0 +1,56 @@
...
+          exp = (Math.log2(bytes) / 10.0).floor
+          converted_number = (100.0 * (bytes / 1024.0**exp)).round / 100.0
+
+          if (converted_number - 1024.0).abs < Float::EPSILON
+            exp += 1
+            converted_number = 1.00
+          end
+          { exp: exp, converted_number: converted_number }
```
sebastian-miclea, line 51: You could return two values, I don't see a need for a hash.

lib/facts_utils/converter.rb
oanatmaria: Done

lib/facts_utils/converter.rb
oanatmaria: Done

lib/facts_utils/converter.rb
oanatmaria: Done

lib/facts_utils/converter.rb
oanatmaria: Done

lib/facts_utils/converter.rb
oanatmaria: Done

lib/facts_utils/converter.rb
oanatmaria: I kept the _readable because i think it is easier to understand what that method is supposed to do.

sebastian-miclea (dismissed):

BogdanIrimie (approved):

PR #455, opened 2020-04-15
(FACT-2559) Fix Facter.debugging? call when Facter not fully loaded

Filipovici-Andrei:


lib/framework/logging/logger.rb
```diff
@@ -94,5 +94,15 @@ def error(msg, colorize = false)
...
     end
+
+    private
+
+    def debugging_active?
+      if Facter.respond_to?(:debugging?)
+        Facter.debugging?
+      else
+        true
+      end
```
gimmyxd, line 105: this can be written as 
```ruby
Facter.respond_to?(:debugging?) && Facter.debugging?
```

spec/framework/logging/logger_spec.rb
```diff
@@ -3,91 +3,123 @@
...
 
   describe '#debug' do
     before do
       allow(Facter).to receive(:debugging?).and_return(true)
     end
 
-    let(:handler) { instance_spy(Logger) }
-
-    it 'noops of debugging is not set' do
+    it 'no debug messages are sent if debugging is set to false' do
```
gimmyxd, line 17: usually `if` statements from example should be a context

```ruby
context 'when debugging is false' do
	it 'does not send a message'
	end
end
``` 

http://www.betterspecs.org/#contexts

spec/framework/logging/logger_spec.rb
```diff
@@ -3,91 +3,123 @@
...
+        expect(multi_logger_double).to have_received(:debug).with('Class - debug_message')
+      end
+    end
+
+    it_behaves_like 'writes debug message'
+
+    context 'when message callback is provided' do
+      after do
+        Facter::Log.class_variable_set(:@@message_callback, nil)
+      end
```
gimmyxd, line 50: Why is this needed? is the logger keeping `messasge_callback` between calls?

spec/framework/logging/logger_spec.rb
Filipovici-Andrei: done

lib/framework/logging/logger.rb
Filipovici-Andrei: done

spec/framework/logging/logger_spec.rb
Filipovici-Andrei: Yes, the logger is keeping the message_callback between calls

lib/framework/logging/logger.rb
Filipovici-Andrei: yes, you are correct

spec/framework/logging/logger_spec.rb
Filipovici-Andrei: I'm not sure if this is the expected behaviour

sebastian-miclea (dismissed):

BogdanIrimie (approved):

PR #456, opened 2020-04-16
(FACT-2553) remove double backslashes from windows path

oanatmaria:
Description of the problem: On Windows facts that contains paths (like path and system32 fact) are double backslashed when facter is executed without any formatter. When facter is executed with --yaml argument the same facts (path and system32) are single backslashed instead of double backslashed.


lib/framework/formatters/legacy_fact_formatter.rb
```diff
@@ -65,7 +65,8 @@ def hash_to_facter_format(facts_hash)
       @log.debug('Remove quotes from parent nodes')
       pretty_json.gsub!(/\"(.*)\"\ =>/, '\1 =>')
 
-      pretty_json
+      @log.debug('Remove double backslashes from paths')
```
glennsarti, line 68: This is dangerous.  UNC paths use double backslashes.

spec/framework/formatters/yaml_fact_formatter_spec.rb
```diff
@@ -1,52 +1,65 @@
...
+      instance_spy(Facter::ResolvedFact, name: 'path', value: value,
+                                         user_query: '', filter_tokens: [], type: :core)
+    end
+    let(:value) { 'C:\\Program Files\\Puppet Labs\\Puppet\\bin;C:\\cygwin64\\bin' }
+    let(:expected_output) { "path: C:\\\\Program Files\\\\Puppet Labs\\\\Puppet\\\\bin;C:\\\\cygwin64\\\\bin\n" }
+    let(:resolved_fact_list) { [win_path] }
 
-    expect(formatted_output).to eq(expected_output)
+    it 'formats path with double escaped backslashes' do
+      expect(yaml_formatter.format(resolved_fact_list)).to eq(expected_output)
```
joshcooper, line 62: I would expect the resolved fact to contain a string with single backslashes (like `'C:\Program Files\...'`), and have the formatter preserve that:

```
irb(main):003:0> puts YAML.dump('C:\Program')
--- C:\Program
```

lib/framework/formatters/legacy_fact_formatter.rb
oanatmaria: After JSON.generate all backslashes are doubled, this means that UNC paths now have 4 backslashes. I simply want to revert that by replacing 2 backslashes with one, resulting, in this case, that UNC paths will be 2 backslashes.

spec/framework/formatters/yaml_fact_formatter_spec.rb
oanatmaria: What you described is exactly the behaviour of Facter 4. When facter-ng is executed with '--yaml' backslashes are not doubled, the need to double them came from the fact that facter 3 prints doubled backslashes when is executed with '--yaml'

BogdanIrimie (approved):

PR #458, opened 2020-04-16
(FACT-2562) Correctly load custom and external fact directories

BogdanIrimie:


spec/facter/facter_spec.rb
```diff
@@ -286,7 +286,7 @@
 
   describe '#search_path' do
     it 'sends call to LegacyFacter' do
```
glennsarti, line 288: Test description is now wrong

spec/facter/facter_spec.rb
```diff
@@ -302,7 +302,7 @@
 
   describe '#search_external_path' do
     it 'sends call to LegacyFacter' do
```
glennsarti, line 304: Test description is now wrong

spec/facter/facter_spec.rb
IrimieBogdan: Updated the test description

spec/facter/facter_spec.rb
IrimieBogdan: Updated the test description

sebastian-miclea (approved):

Filipovici-Andrei (approved):

PR #459, opened 2020-04-20
(FACT-2550) Fix for FileHelper unit tests

Filipovici-Andrei:
Instead of setting a mocked logger on the FileHelper class, I mock the multi logger inside the Log class.


lib/util/file_helper.rb
```diff
@@ -4,18 +4,26 @@ module Facter
...
       @log = Log.new(self)
+
       class << self
         DEBUG_MESSAGE = 'File at: %s is not accessible.'
 
         def safe_read(path, result_if_not_readable = '')
           return File.read(path) if File.readable?(path)
 
-          result_if_not_readable
+          failed_to_read(path, result_if_not_readable)
```
sebastian-miclea, line 14: Could you move **result_if_not_readable** after the call to failed_to_read? I see no use in sending it trough the method.

lib/util/file_helper.rb
```diff
@@ -4,18 +4,26 @@ module Facter
   module Util
     class FileHelper
       @log = Log.new(self)
+
       class << self
         DEBUG_MESSAGE = 'File at: %s is not accessible.'
 
         def safe_read(path, result_if_not_readable = '')
```
sebastian-miclea, line 11: I would rename ** result_if_not_readable** to **default_return**

lib/util/file_helper.rb
```diff
@@ -4,18 +4,26 @@ module Facter
...
 
         def safe_readlines(path, result_if_not_readable = [])
           return File.readlines(path) if File.readable?(path)
 
+          failed_to_read(path, result_if_not_readable)
+        end
+
+        private
+
+        def failed_to_read(path, result_if_not_readable)
```
sebastian-miclea, line 25: Can we rename this to **log_failed_read** or something similar?

PR #460, opened 2020-04-20
Add support for *BSD load averages

smortex:
The *BSD osfamily has `getloadavg(3)` to retrieve the load averages.  This implementation was based on the Aix implementation (which also use FFI), and was tested on FreeBSD.

It is expected to work likewise on OpenBSD and NetBSD (and other derivative) since the getloadavg() function appeared in 4.3BSD-Reno.


BogdanIrimie (approved):

IrimieBogdan:
@smortex the PR looks good!

PR #461, opened 2020-04-20
Fix os.family fact on *BSD

smortex:
On *BSD rely on uname to determine os.family (just as the Aix does).  The current behavior rely on the Solaris fact that hardcodes "Solaris" as the os.family.


IrimieBogdan:
@smortex some PR's are just perfect, and this is one of them :)

BogdanIrimie (approved):

PR #466, opened 2020-04-22
(FACT-2565) Debian development versions causes fatal error when resolving os.release

Filipovici-Andrei:


lib/facts/debian/os/distro/release.rb
```diff
@@ -13,7 +13,8 @@ def call_the_resolver
             return Facter::ResolvedFact.new(FACT_NAME, nil) unless fact_value
 
             versions = fact_value.split('.')
-            release = { 'full' => fact_value, 'major' => versions[0], 'minor' => versions[1].gsub(/^0([1-9])/, '\1') }
+            release = { 'full' => fact_value, 'major' => versions[0] }
+            release['minor'] = versions[1].gsub(/^0([1-9])/, '\1') unless versions[1].nil?
```
oanatmaria, line 17: if versions[1] to avoid nil check

Filipovici-Andrei: yes

lib/facts/debian/os/distro/release.rb
```diff
@@ -13,8 +13,8 @@ def call_the_resolver
             return Facter::ResolvedFact.new(FACT_NAME, nil) unless fact_value
 
             versions = fact_value.split('.')
-            release = { 'full' => fact_value, 'major' => versions[0], 'minor' => versions[1].gsub(/^0([1-9])/, '\1') }
-
+            release = { 'full' => fact_value, 'major' => versions[0] }
+            release['minor'] = versions[1].gsub(/^0([1-9])/, '\1') if versions[1]
```
IrimieBogdan, line 17: We should rewrite to
```
            release = { }
            release['full'] = fact_value
            release['major'] = versions[0]
            release['minor'] = versions[1].gsub(/^0([1-9])/, '\1') if versions[1]
```
By doing this rewrite it will be easier to break this fact in 3 facts and allow individual blocking and caching.

lib/facts/debian/os/release.rb
```diff
@@ -13,11 +13,8 @@ def call_the_resolver
           return Facter::ResolvedFact.new(FACT_NAME, nil) unless fact_value
 
           versions = fact_value.split('.')
-          release = {
-            'full' => fact_value,
-            'major' => versions[0],
-            'minor' => versions[1].gsub(/^0([1-9])/, '\1')
-          }
+          release = { 'full' => fact_value, 'major' => versions[0] }
+          release['minor'] = versions[1].gsub(/^0([1-9])/, '\1') if versions[1]
```
IrimieBogdan, line 17: Change similar to `release.rb`

BogdanIrimie (dismissed):

BogdanIrimie (approved):

lib/framework/core/fact/internal/internal_fact_manager.rb
```diff
@@ -18,14 +20,18 @@ def filter_internal_facts(searched_facts)
...
         threads << Thread.new do
-          fact = CoreFact.new(searched_fact)
-          fact.create
+          begin
+            fact = CoreFact.new(searched_fact)
+            fact.create
+          rescue StandardError => e
+            @@log.error(e.backtrace.join("\n"))
+            nil
+          end
```
gimmyxd, line 34: This will catch all the errors, even dev errors and unhandled cases even from resolvers and will just output a backtrace. I don't think this is the way to go for this situation. 
If i call a method on nil in the resolver, this part will catch it for example and return nil, passing nil around and generating more errors.

PR #471, opened 2020-04-28
(FACT-2553) Quote special string in YAML format

oanatmaria:


lib/framework/formatters/yaml_fact_formatter.rb
```diff
@@ -6,13 +6,58 @@ def initialize
...
+      user_queries = resolved_facts.uniq(&:user_query).map(&:user_query)
+
+      facts_to_display = if user_queries.count == 1 && user_queries.first.empty?
+                           FormatterHelper.retrieve_fact_collection(resolved_facts)
+                         else
+                           FormatterHelper.retrieve_facts_to_display_for_user_query(user_queries, resolved_facts)
+                         end
+
+      facts_to_display = Psych.parse_stream(facts_to_display.to_yaml)
+      yaml_pretty = quote_special_strings(facts_to_display)
```
IrimieBogdan, line 19: The `tag_directives` from `Psych::Nodes::Document` get strange values in JRuby, we should set them manually after creating the `Psych::Nodes::Stream` 

```
# set tag_directives manually because they have strange values (%TAG !! tag:yaml.org,2002:) in JRuby.
facts_to_display.children[0].tag_directives = []
```

BogdanIrimie (approved):

PR #477, opened 2020-04-28
(FACT-2569) Run acceptance tests on Ubuntu GitHub actions

Filipovici-Andrei:


.github/actions/presuite.sh
```diff
@@ -0,0 +1,42 @@
...
+
+cd $cwd/$FACTER_4_ROOT
+puppet_gem_command=/opt/puppetlabs/puppet/bin/gem
+$puppet_gem_command build agent/facter-ng.gemspec
+$puppet_gem_command uninstall facter-ng
+$puppet_gem_command install -f facter-ng-*.gem
+
+cd /opt/puppetlabs/puppet/bin
+mv facter-ng facter
+
```
IrimieBogdan, line 39: Would be great if you could add

```
printf "\nFacter version\n\n"
facter --version

printf "\Puppet facts facter version\n\n"
puppet facts | grep facterversion
```

BogdanIrimie (dismissed):

.github/actions/presuite.sh
Filipovici-Andrei: yes, sure

BogdanIrimie (approved):

PR #483, opened 2020-04-29
(FACT-2235) Add Aix processors fact

oanatmaria:


lib/resolvers/aix/processors.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    module Aix
+      class Processors < BaseResolver
+        @log = Facter::Log.new(self)
```
IrimieBogdan, line 7: We should use the logger from `BaseResolver`

oanatmaria: Done

BogdanIrimie (approved):

PR #487, opened 2020-04-30
(FACT-2420) partial fix for verify_partial_doubles

gimmyxd:
The Windows and FFI helper specs are mocking FFI classes and extensions which does not exists because we monkey patch `Kernel.require` to not load FFI stuff. Until that part is fixed we cannot enable `verify_partial_doubles`.

### DIFF
--- a/lib/custom_facts/core/execution.rb
+++ b/lib/custom_facts/core/execution.rb
@@ -51,7 +51,14 @@ module Facter
       # @param path [String] the path to check
       # @param platform [:posix,:windows,nil] the platform logic to use
       def absolute_path?(path, platform = nil)
-        @@impl.absolute_path?(path, platform)
+        case platform
+        when :posix
+          Facter::Core::Execution::Posix.new.absolute_path?(path)
+        when :windows
+          Facter::Core::Execution::Windows.new.absolute_path?(path)
+        else
+          @@impl.absolute_path?(path)
+        end
       end
 
       # Given a command line, this returns the command line with the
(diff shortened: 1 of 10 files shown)


lib/custom_facts/core/execution.rb
```diff
@@ -51,7 +51,14 @@ def which(bin)
...
       def absolute_path?(path, platform = nil)
-        @@impl.absolute_path?(path, platform)
+        case platform
+        when :posix
+          Facter::Core::Execution::Posix.new.absolute_path?(path)
+        when :windows
+          Facter::Core::Execution::Windows.new.absolute_path?(path)
+        else
+          @@impl.absolute_path?(path)
+        end
```
gimmyxd, line 61: This was actually a bug swallowed by RSpec, as the `@@impl.absolute_path?` was never called and it passed by that this method is not even accepting two parameters on Windows and Posix implementations

spec/framework/core/fact_loaders/external_fact_loader_spec.rb
```diff
@@ -38,20 +38,6 @@
...
-      end
-
-      it 'does not load custom facts (does not call LegacyFacter.search)' do
-        allow(Facter::Options).to receive(:custom_dir?).and_return(false)
-        expect(LegacyFacter).not_to receive(:search)
-
-        external_fact_loader = Facter::ExternalFactLoader.new
-        external_fact_loader.custom_facts
-      end
-    end
```
gimmyxd, line 54: This was not testing anything relevant anymore

BogdanIrimie (approved):

PR #485, opened 2020-04-30
Add os.release facts on FreeBSD

smortex:
This PR provide support for Facter 3.x `os.release` facts on FreeBSD to Facter 4:

Before:
```sh-session
romain@zappy ~/Projects/facter-ng % bundle exec bin/facter os        
cat: /etc/release: No such file or directory
[2020-04-29 18:51:23.841325 ] ERROR Facter::Resolvers::SolarisRelease - Could not build release fact because of missing or empty file /etc/release 
cat: /etc/release: No such file or directory
[2020-04-29 18:51:23.842256 ] ERROR Facter::Resolvers::SolarisRelease - Could not build release fact because of missing or empty file /etc/release 
cat: /etc/release: No such file or directory
[2020-04-29 18:51:23.843051 ] ERROR Facter::Resolvers::SolarisRelease - Could not build release fact because of missing or empty file /etc/release 
{
  architecture => "amd64",
  family => "FreeBSD",
  hardware => "amd64",
  name => "FreeBSD",
  release => {
    full => null,
    major => null,
    minor => null
  }
}
```

After:
```sh-session
romain@zappy ~/Projects/facter-ng % bundle exec bin/facter os      
{
  architecture => "amd64",
  family => "FreeBSD",
  hardware => "amd64",
  name => "FreeBSD",
  release => {
    branch => "RELEASE-p4",
    full => "12.1-RELEASE-p4",
    major => "12",
    minor => "1",
    patchlevel => "4"
  }
}
```

Compare to:

```sh-session
romain@zappy ~/Projects/facter-ng % facter os
{
  architecture => "amd64",
  family => "FreeBSD",
  hardware => "amd64",
  name => "FreeBSD",
  release => {
    branch => "RELEASE-p4",
    full => "12.1-RELEASE-p4",
    major => "12",
    minor => "1",
    patchlevel => "4"
  }
}
```

PR #489, opened 2020-05-01
Add support for processors facts on *BSD

smortex:
Add a basic sysctl(3) FFI wrapper for getting processors information. This allows Facter 4 to return the same processors facts as returned by Facter 3 on FreeBSD.

While working on this I spotted a typo in an unrelated file, so fixed it in another commit.


smortex:
I might be moving too fast…  this depends of bits in #485  :roll_eyes:

lib/resolvers/freebsd/ffi/ffi_helper.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'ffi'
+
+module Facter
+  module Freebsd
+    module FfiHelper
```
IrimieBogdan, line 7: Please add some tests for `Freebsd :: FfiHelper `

lib/resolvers/freebsd/processors.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require 'resolvers/freebsd/ffi/ffi_helper'
```
IrimieBogdan, line 3: I would build an absolute path for require e.g. `require "#{ROOT_DIR}/lib/resolvers//freebsd/ffi/ffi_helper"` and move the require inside the method where it is used. We have a mechanism in [BaseResolver](https://github.com/puppetlabs/facter-ng/blob/91df2922b84701dbce42ff77147af46398fdab4f/lib/resolvers/base_resolver.rb#L23) that catches load exceptions. For example if we run Facter 4 on FreeBSD without ffi, it will still work for the facts that can be resolved without ffi.

smortex: Done!

lib/resolvers/freebsd/ffi/ffi_helper.rb
smortex: Did it, but it raise a few concerns for me because these tests just mock everything and we are testing the mock framework more than anything else :cry:

smortex:
I realized that most of the code could work on all *BSD systems with just a few tweaks, so I refactored the PR a bit to more the code from the FreeBSD to the BSD directories.

Yet, some differences exist for some aspects of the processors fact, and I could not manage yet to do things correctly.  The BSD and FreeBSD resolvers are almost the same, only the way of fetching the CPU speed is different, and I would like to avoid duplicating all the code in this class.  I though about inheritance, but it is not really working because both classes are instanced and the common facts are fetched twice :-/

Filipovici-Andrei:
We had some problems with the GitHub CI, but now it's fixed. So a merge with master will fix it for your PR as well.

smortex:
Inheritance in resolvers not being really helpful, I completely separated the Bsd resolver and the Freebsd one.  That way, facts are not collected twice by the Bsd and the Freebsd resolvers on FreeBSD as it was previously the case.

I reabased the work on top of master, and since the commit where going back and forth, I squashed them to make the change more readable.  Please review, thanks!

IrimieBogdan:
@smortex please rebase on master to fix  RuboCop. Regarding `Checks / coverage` there is a known issue with PRs from Forks outside organisation. If all other test pass except `Checks / coverage` the PR looks good and we will merge it.

BogdanIrimie (approved):

PR #500, opened 2020-05-11
(FACT-2531) Fix for tests/facts/validate_file_system_size_bytes.rb

oanatmaria:
Description of the problem:
 - for linux mountpoints all values have to be positive
- architecture fact for 32-bit os must be i386
- for rhel os release fact must be split 6.2 => major 6, minor 2
- mountpoints for aix contains an additional wrong fact : "mounted"=>{:device=>"node", :filesystem=>"mounted", :options=>["options"]} 
- physical count for processors fact is 0 when "physical id" is not present in /proc/cpuinfo

Description of the fix:
- call abs method for mountpoints bytes facts
- force architecture value to 'i386' if it matches /i[3456]86|pentium/
- implement os.release for rhel
- improve mountpoints regex to exclude file's header
- when physical count for processors fact is 0, count subdirectories from /sys/devices/system/cpu, that contains cpu(number) in name

This PR solves: tests/facts/validate_file_system_size_bytes.rb  test for all oses that have mountpoints fact and os_processors_and_kernel test **for all** oses.

### DIFF
--- a/lib/facts/debian/architecture.rb
+++ b/lib/facts/debian/architecture.rb
@@ -10,6 +10,7 @@ module Facts
         def call_the_resolver
           fact_value = Facter::Resolvers::Uname.resolve(:machine)
           fact_value = 'amd64' if fact_value == 'x86_64'
+          fact_value = 'i386' if fact_value =~ /i[3456]86|pentium/
 
           [Facter::ResolvedFact.new(FACT_NAME, fact_value), Facter::ResolvedFact.new(ALIASES, fact_value, :legacy)]
         end
--- a/lib/facts/linux/os/architecture.rb
+++ b/lib/facts/linux/os/architecture.rb
@@ -10,6 +10,8 @@ module Facts
         def call_the_resolver
           fact_value = Facter::Resolvers::Uname.resolve(:machine)
 
+          fact_value = 'i386' if fact_value =~ /i[3456]86|pentium/
+
           [Facter::ResolvedFact.new(FACT_NAME, fact_value), Facter::ResolvedFact.new(ALIASES, fact_value, :legacy)]
         end
       end
--- a/lib/resolvers/aix/mountpoints.rb
+++ b/lib/resolvers/aix/mountpoints.rb
@@ -19,7 +19,7 @@ module Facter
             @fact_list[:mountpoints] = {}
             output = Facter::Core::Execution.execute('mount', logger: log)
             output.split("\n").map do |line|
-              next if line =~ /\snode\s|---|procfs|ahafs/
+              next if line =~ /node|---|procfs|ahafs/
 
               elem = line.split("\s")
 
--- a/lib/resolvers/processors_resolver.rb
+++ b/lib/resolvers/processors_resolver.rb
@@ -25,6 +25,7 @@ module Facter
             read_processors(cpuinfo_output) # + model names
 
             @fact_list[:physical_count] = @fact_list[:physical_processors].uniq.length
+            @fact_list[:physical_count] = physical_devices_count if @fact_list[:physical_count].zero?
             @fact_list[fact_name]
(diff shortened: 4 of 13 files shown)


lib/resolvers/processors_resolver.rb
```diff
@@ -51,6 +52,12 @@ def construct_models_list(tokens)
           def count_physical_processors(tokens)
             @fact_list[:physical_processors] << tokens.last.strip.to_i if tokens.first.strip == 'physical id'
           end
+
+          def search_in_system_devices
+            dirs = Dir.entries('/sys/devices/system/cpu').reject { |dir| dir =~ /\.+/ }
+
+            dirs.count { |dir| dir =~ /cpu[0-9]+$/ }
```
gimmyxd, line 59: I think you can use count directly, as you are passing a block and counting specific entries:
```ruby
Dir.entries('/sys/devices/system/cpu').count { |dir| dir =~ /cpu[0-9]+$/ }
```

Also maybe the method name can be more suggestive: `physical_devices_count` or something like that.

lib/facts/debian/architecture.rb
```diff
@@ -10,6 +10,7 @@ class Architecture
         def call_the_resolver
           fact_value = Facter::Resolvers::Uname.resolve(:machine)
           fact_value = 'amd64' if fact_value == 'x86_64'
+          fact_value = fact_value =~ /i[3456]86|pentium/ ? 'i386' : fact_value
```
gimmyxd, line 13:
```ruby
fact_value = i386 if `fact_value =~ /i[3456]86|pentium/
```
You already have a good value in `fact_value` change it only if you need to

lib/facts/linux/os/architecture.rb
```diff
@@ -10,6 +10,8 @@ class Architecture
         def call_the_resolver
           fact_value = Facter::Resolvers::Uname.resolve(:machine)
 
+          fact_value = fact_value =~ /i[3456]86|pentium/ ? 'i386' : fact_value
```
gimmyxd, line 13: Same as on lib/facts/debian/architecture.rb

lib/resolvers/aix/mountpoints.rb
```diff
@@ -19,7 +19,7 @@ def read_mount(fact_name)
             @fact_list[:mountpoints] = {}
             output = Facter::Core::Execution.execute('mount', logger: log)
             output.split("\n").map do |line|
-              next if line =~ /\snode\s|---|procfs|ahafs/
+              next if line =~ /node|---|procfs|ahafs/
```
gimmyxd, line 22: Is this change covered in tests? i don't see aix/mountpoints_spec.rb to have changed

gimmyxd:
Would be useful to describe in the commit message what this fixes, as the PR title states is fixing an acceptance tests but is not clear what was the issue.

lib/resolvers/processors_resolver.rb
oanatmaria: done

lib/facts/debian/architecture.rb
oanatmaria: done

lib/facts/linux/os/architecture.rb
oanatmaria: done

lib/resolvers/aix/mountpoints.rb
oanatmaria: done

oanatmaria: the change is in fixtures/mount

gimmyxd (approved):

BogdanIrimie (approved):

Filipovici-Andrei (approved):

PR #501, opened 2020-05-12
(FACT-2613) Migrate unit tests to github actions

BogdanIrimie:


.github/workflows/unit_tests.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+on: [pull_request]
+
+jobs:
+  unit_tests:
+    name: Run unit tests on Ruby
+    strategy:
+      matrix:
+        ruby: [ 2.3, 2.4, 2.5, 2.6, 2.7, jruby ]
+    runs-on: ubuntu-18.04
```
gimmyxd, line 12: might be useful to also run them on windows?

.github/workflows/unit_tests.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Checkout current PR
+        uses: actions/checkout@v2
+
+      - name: Rubocop & rspec checks
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - run: gem install bundler
+      - run: bundle update
+      - run: bundle exec rubocop
```
gimmyxd, line 24: i don't think we need to run rubocop on each ruby version. Maybe have a separate workflow for this?

gimmyxd: Also check commits is missing, maybe that can also be another step/workflow

.github/workflows/unit_tests.yaml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - name: Checkout current PR
+        uses: actions/checkout@v2
+
+      - name: Rspec checks
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - run: gem install bundler
+      - run: bundle update
```
gimmyxd, line 23: bundle install --jobs 3 --retry 3 would be faster

.github/workflows/unit_tests.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Checkout current PR
+        uses: actions/checkout@v2
+
+      - name: Rubocop checks
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.6
+      - run: gem install bundler
+      - run: bundle update
+      - run: bundle exec rubocop
```
gimmyxd, line 38: `rubocop --parallel` to make it faster

.github/workflows/unit_tests.yaml
```diff
@@ -0,0 +1,38 @@
+---
+name: Unit tests
+
+on: [pull_request]
+
+jobs:
```
gimmyxd, line 6: We should consider using https://github.com/actions/cache since gem installing can be time consuming, and if nothing changed in the Gemfile, there is no reason to run bundler.
Eg: https://github.com/puppetlabs/bolt/blob/master/.github/workflows/linux.yaml#L30-L35

.github/workflows/checks.yaml
```diff
@@ -0,0 +1,33 @@
+---
+name: Checks
+
+on: [pull_request]
+
+jobs:
+  rspec_checks:
```
gimmyxd, line 7: `rubocop_checks`

IrimieBogdan: changed

gimmyxd (approved):

PR #507, opened 2020-05-12
(FACT-2533) Fix for tests/facts/partitions.rb

oanatmaria:
Description of the problem: when retrieving partuuid fact for partitions there is an extra space in it's value
Description of the fix: remove extra space from partuuid value

### DIFF
--- a/lib/resolvers/partitions.rb
+++ b/lib/resolvers/partitions.rb
@@ -95,7 +95,9 @@ module Facter
         def execute_and_extract_blkid_info
           stdout = Facter::Core::Execution.execute('blkid', logger: log)
           output_hash = Hash[*stdout.split(/^([^:]+):/)[1..-1]]
-          output_hash.each { |key, value| output_hash[key] = Hash[*value.delete('"').chomp.split(/ ([^= ]+)=/)[1..-1]] }
+          output_hash.each do |key, value|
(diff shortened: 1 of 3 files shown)


lib/resolvers/partitions.rb
```diff
@@ -96,6 +96,7 @@ def execute_and_extract_blkid_info
           stdout = Facter::Core::Execution.execute('blkid', logger: log)
           output_hash = Hash[*stdout.split(/^([^:]+):/)[1..-1]]
           output_hash.each { |key, value| output_hash[key] = Hash[*value.delete('"').chomp.split(/ ([^= ]+)=/)[1..-1]] }
+          output_hash.each { |_key, value| value['PARTUUID'] = value['PARTUUID'].gsub(/\s/, '') if value['PARTUUID'] }
```
gimmyxd, line 99: using https://ruby-doc.org/core-2.7.1/String.html#method-i-rstrip might solve the issue in a single iteration since the problem seems to be caused by the way we are splitting the value and also taking the last extra whitespace 

```ruby
output_hash.each do |key, value|
  output_hash[key] = Hash[*value.delete('"').chomp.rstrip.split(/ ([^= ]+)=/)[1..-1]]
end
``

oanatmaria: Done

gimmyxd (approved):

PR #520, opened 2020-05-15
(FACT-2602) Add docker/Lxc resolver for Linux

oanatmaria:


lib/resolvers/docker_lxc.rb
```diff
@@ -0,0 +1,49 @@
...
+        private
+
+        def post_resolve(fact_name)
+          @fact_list.fetch(fact_name) { read_cgroup(fact_name) }
+        end
+
+        def read_cgroup(fact_name)
+          output_cgroup = Util::FileHelper.safe_read('/proc/1/cgroup', nil)
+          output_environ = Util::FileHelper.safe_read('/proc/1/environ', nil)
+          return if !output_cgroup || !output_environ
```
IrimieBogdan, line 23: if output_cgroup.nil? || output_environ.nil?

lib/facts/linux/hypervisors.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Facts
+  module Linux
+    class Hypervisors
+      FACT_NAME = 'hypervisors'
+
+      def call_the_resolver
+        fact_value = check_docker_lxc
```
IrimieBogdan, line 9: Maybe we should remove this method, it's body is just a method call (calls the resolver)

lib/resolvers/docker_lxc.rb
oanatmaria: done

BogdanIrimie (approved):

PR #521, opened 2020-05-18
(FACT-2603) Detect virtual on GCE vms

oanatmaria:


lib/facts/linux/virtual.rb
```diff
@@ -6,9 +6,20 @@ class Virtual
...
-        fact_value = Facter::Resolvers::DockerLxc.resolve(:vm)
+        fact_value = check_docker_lxc || check_gce
         Facter::ResolvedFact.new(FACT_NAME, fact_value)
       end
+
+      def check_gce
+        bios_vendor = Facter::Resolvers::Linux::DmiBios.resolve(:bios_vendor)
+        return 'gce' if bios_vendor&.include?('Google')
+
+        nil
```
gimmyxd, line 17: by defult if the `return 'gce'` is not evaluated `check_gce`, will return `nil`. So you can drop the extra `nil`

oanatmaria: done

BogdanIrimie (approved):

PR #523, opened 2020-05-19
(FACT-2604) Add virt-what resolver

oanatmaria:


lib/resolvers/virt_what.rb
```diff
@@ -0,0 +1,64 @@
...
+          return retrieve_vserver if other_vm =~ /linux_vserver/
+
+          other_vm
+        end
+
+        def retrieve_vserver
+          proc_status_content = Facter::Util::FileHelper.safe_readlines('/proc/self/status', nil)
+          return unless proc_status_content
+
+          proc_status_content.map do |line|
```
IrimieBogdan, line 51: We should use `each` instead of `map`. Map is used when we want to collect the results (e.g. make some changes on all the elements in a list)

lib/resolvers/virt_what.rb
```diff
@@ -0,0 +1,64 @@
...
+            parts = line.split("\s")
+            next unless parts.size.equal?(2)
+
+            next unless parts[0] =~ /^s_context:|^VxID:/
+            return @fact_list[:vserver] = 'vserver_host' if parts[1] =~ /^0$/
```
IrimieBogdan, line 56: The regex might not be necessary here, this can be replaced with `parts[1] == '0'`

lib/resolvers/virt_what.rb
oanatmaria: done

lib/resolvers/virt_what.rb
oanatmaria: done

BogdanIrimie (approved):

PR #528, opened 2020-05-21
(FACT-2632) Log error message if we encounter exceptions while loading custom facts files

BogdanIrimie:


lib/custom_facts/util/loader.rb
```diff
@@ -123,6 +123,8 @@ def load_file(file)
       # @return [Boolean]
       def kernel_load(file)
         Kernel.load(file)
+      rescue StandardError => e
+        Facter.log_exception(e, "error while resolving custom facts in #{file} #{e.message}")
```
gimmyxd, line 127: Something similar seems to be happening for ScriptError: https://github.com/puppetlabs/facter-ng/pull/528/files#diff-088677df2f572654d2a895f4cd71dc81R111-R116

Maybe we can unify this behaviour?

IrimieBogdan: Moved rescue to `load_file`

gimmyxd (approved):

PR #530, opened 2020-05-22
(FACT-2633) Relax rubocop dependancy

oanatmaria:
- Fix new cops offences
- Ruby 2.3 is not supported by newer versions of rubocop

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -14,7 +14,7 @@ jobs:
     name: Ruby version
     strategy:
       matrix:
-        ruby: [ 2.3, 2.7, jruby ]
+        ruby: [ 2.4, 2.7, jruby ]
     runs-on: ubuntu-18.04
     steps:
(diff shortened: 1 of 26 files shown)


gimmyxd:
Making this changes means that Puppet as a gem will not be able to run on Ruby 2.3(due to the missing methods like `match?`) 
- https://apidock.com/ruby/v2_4_6/String/match%3F

- https://github.com/puppetlabs/puppet/blob/master/.gemspec#L21

I think also PDK is using older versions of Ruby.

oanatmaria:
@gimmyxd yes, you are right. This PR won't be merged (marked with do not merge) until we drop support for ruby 2.3 in puppet.

joshcooper:
I filed https://tickets.puppetlabs.com/browse/PUP-10509 to drop old ruby versions in Puppet Platform 7. Until then maybe target ruby 2.3 language features and disable the cop?

IrimieBogdan:
We will continue to support Ruby 2.3

PR #536, opened 2020-05-26
(FACT-2635) Incorrect output for non existing fact

BogdanIrimie:
When querying  for two facts, one that exists and one that does not, Facter 4 outputs only the fact that exists. It should output both facts, the one that does not exist with nil (or empty depending on the formatter) value

e.g. no fix
```
facter os.name test

Darwin
```

e.g. with fix
```
facter os.name test

os.name => Darwin
test =>
```


lib/framework/formatters/legacy_fact_formatter.rb
```diff
@@ -60,7 +60,8 @@ def hash_to_facter_format(facts_hash)
       pretty_json = JSON.pretty_generate(facts_hash)
 
       @log.debug('Change key value delimiter from : to =>')
-      pretty_json.gsub!(/^(.*?)(:)/, '\1 =>')
+      # pretty_json.gsub!(/^(.*?)(:)/, '\1 =>')
```
oanatmaria, line 63: remove comment

lib/framework/core/fact/internal/internal_fact_manager.rb
```diff
@@ -18,6 +21,19 @@ def filter_internal_facts(searched_facts)
       searched_facts.select { |searched_fact| %i[core legacy].include? searched_fact.type }
     end
 
+    def filter_nil_facts(searched_facts)
+      searched_facts.select { |searched_fact| %i[nil].include? searched_fact.type }
+    end
+
+    def resolve_nil_facts(searched_facts)
+      resolved_facts = []
+      searched_facts.select { |fact| fact.type == :nil }.each do |fact|
```
gimmyxd, line 30: Why is this filtering again the `nil` type facts? it is called with the result of `filter_nil_facts`

Does `filter_nil_facts` is not really useful as it is called only once.

I would suggest to drop this method or create a `nil_searched_facts` private methods that memoize the result and use it in `resolve_nil_facts`

IrimieBogdan: Grate suggestion 👍  , removed the `filter_nil_facts `

spec/framework/core/fact/internal/internal_fact_manager_spec.rb
```diff
@@ -32,6 +32,24 @@
       expect(resolved_facts).to eq([resolved_fact])
     end
 
+    context 'when resolved fac is of type nil' do
```
oanatmaria, line 35: resolved fact

IrimieBogdan: Fixed

oanatmaria (approved):

PR #537, opened 2020-05-27
(FACT-2636) Set external as fact_type for environment variable facts.

BogdanIrimie:
Environment variable external facts were missing `fact_type: :external` option, and because of this they were ignored when resolving facts. This is required since `Facter 4.0.22` as external facts are identified based on `fact_type ` option.

### DIFF
--- a/lib/custom_facts/util/loader.rb
+++ b/lib/custom_facts/util/loader.rb
@@ -141,7 +141,7 @@ module LegacyFacter
           # match it.
           next if fact && (env_name != fact)
 
-          LegacyFacter.add(Regexp.last_match(1)) do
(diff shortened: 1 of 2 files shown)


Magisus:
Would it be possible to get this released soon? PDB tests are still failing because of it. If not, please let me know, so I can revert our switch to testing with Facter 4 in PDB and puppetserver.

Magisus (approved):
I verified that this fix resolves the issue, using the repro case in the ticket.

IrimieBogdan:
@Magisus released [Facter 4.0.25](https://rubygems.org/gems/facter/versions/4.0.25) today, please give it a try.

PR #543, opened 2020-06-04
FACT-2650 Fix bug when loading external facts

Filipovici-Andrei:
When running facter with external facts, in powershell, the paths given to the search_path method are not interpreted correctly and no facts are resolved.


spec/custom_facts/util/loader_spec.rb
```diff
@@ -280,6 +280,22 @@ def loader_from(places)
...
+        allow(Dir).to receive(:glob).with('/one/dir/*.rb').and_return %w[/one/dir/a.rb]
+      end
+
+      dir_paths = ['//one///dir', '//one///\\dir', '/one///\/\dir', '\one///\\dir']
+
+      dir_paths.each do |dir_path|
+        it 'corrects the directory path' do
+          allow(loader).to receive(:search_path).and_return [dir_path]
+
+          loader.load_all
```
IrimieBogdan, line 295: There is no expect

BogdanIrimie (approved):

PR #544, opened 2020-06-04
(FACT-2620) Add EC2 facts for linux

oanatmaria:


lib/resolvers/ec2.rb
```diff
@@ -0,0 +1,67 @@
...
+        def get_data_from(url)
+          parsed_url = URI.parse(url)
+          http = Net::HTTP.new(parsed_url.host)
+
+          session_timeout = ENV['EC2_SESSION_TIMEOUT']
+          http.read_timeout = session_timeout ? session_timeout.to_i : EC2_SESSION_TIMEOUT
+          http.open_timeout = EC2_CONNECTION_TIMEOUT
+          resp = http.get(parsed_url.path)
+
+          resp.code.to_i.equal?(200) ? resp.body : ''
```
Filipovici-Andrei, line 59: this line can be moved inside an ensure block and delete the ' ' after log in rescue

lib/resolvers/ec2.rb
```diff
@@ -0,0 +1,67 @@
...
+          end
+        end
+
+        def get_data_from(url)
+          parsed_url = URI.parse(url)
+          http = Net::HTTP.new(parsed_url.host)
+
+          session_timeout = ENV['EC2_SESSION_TIMEOUT']
+          http.read_timeout = session_timeout ? session_timeout.to_i : EC2_SESSION_TIMEOUT
+          http.open_timeout = EC2_CONNECTION_TIMEOUT
```
Filipovici-Andrei, line 56: this can be moved in a method called config_http

lib/resolvers/ec2.rb
```diff
@@ -0,0 +1,67 @@
...
+            next if name == 'security-credentials/'
+
+            if !name.end_with?('/')
+              container[name] = get_data_from("#{url}#{name}").strip
+            else
+              child = {}
+              child[name] = query_for_metadata("#{url}#{name}", child)
+              child.reject! { |key, _info| key == name }
+              name = name.gsub(%r{/$}, '')
+              container[name] = child
```
Filipovici-Andrei, line 45: this can be moved in a method called name_parsing or something better. Even name should be changed into something more suggestive

lib/resolvers/ec2.rb
```diff
@@ -0,0 +1,67 @@
...
+            end
+          end
+        end
+
+        def get_data_from(url)
+          parsed_url = URI.parse(url)
+          http = Net::HTTP.new(parsed_url.host)
+
+          session_timeout = ENV['EC2_SESSION_TIMEOUT']
+          http.read_timeout = session_timeout ? session_timeout.to_i : EC2_SESSION_TIMEOUT
```
sebastian-miclea, line 55: This condition can be moved where the constant is declared.`EC2_SESSION_TIMEOUT = ENV['k'] || 5`
Or, create a method that returns the timeout and will also handle the to_i part.

lib/resolvers/ec2.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require 'net/http'
+
+module Facter
+  module Resolvers
+    class Ec2 < BaseResolver
+      @semaphore = Mutex.new
+      @fact_list ||= {}
+      EC2_METADATA_ROOT_URL = 'http://169.254.169.254/latest/meta-data/'
```
sebastian-miclea, line 10: Shouldn't these be configurable somehow?

lib/resolvers/ec2.rb
```diff
@@ -0,0 +1,67 @@
...
+          @fact_list[fact_name]
+        end
+
+        def query_for_metadata(url, container)
+          metadata = get_data_from(url)
+          metadata.each_line do |line|
+            next if line.empty?
+
+            array_match = /^(\d+)=.*$/.match(line)
+            name = array_match && array_match[1] ? "#{array_match[1]}/" : line.strip
```
IrimieBogdan, line 35: is `&& array_match[1]` required? The regex already specifies that there must be at lest one digit before `=` and the digit will be included in the group, so it will be accessible with `array_match[1]`

lib/resolvers/ec2.rb
```diff
@@ -0,0 +1,67 @@
...
+            next if name == 'security-credentials/'
```
IrimieBogdan, line 36: Maybe `name` should be replaced `http_path_element` to make it clear that it is a part of a HTTP path

lib/resolvers/ec2.rb
IrimieBogdan, line 35: You can extract this in a method 'build_path_component(line)'
e.g.
```
def build_path_component(line)
        array_match = /^(\d+)=.*$/.match(line)
        array_match && array_match[1] ? "#{array_match[1]}/" : line.strip
end
```

lib/resolvers/ec2.rb
```diff
@@ -0,0 +1,67 @@
...
+
+            if !name.end_with?('/')
```
IrimieBogdan, line 38: `!name.end_with?('/')` the negation should be dropped and the `if` and `else` blocks switched

e.g.
```
if name.end_with?('/')
  child = {}
  child[name] = query_for_metadata("#{url}#{name}", child)
  child.reject! { |key, _info| key == name }
  name = name.gsub(%r{/$}, '')
  container[name] = child
else
  container[name] = get_data_from("#{url}#{name}").strip
end
```

lib/resolvers/ec2.rb
```diff
@@ -0,0 +1,67 @@
...
+              container[name] = get_data_from("#{url}#{name}").strip
+            else
+              child = {}
+              child[name] = query_for_metadata("#{url}#{name}", child)
+              child.reject! { |key, _info| key == name }
+              name = name.gsub(%r{/$}, '')
```
IrimieBogdan, line 44: The regex could be replaced with `.chomp('/')`

lib/resolvers/ec2.rb
IrimieBogdan: I would go even further and extract all the logic for timeout in a different method e.g.

```
def configure_timout(http)
  session_timeout = ENV['EC2_SESSION_TIMEOUT']
  http.read_timeout = session_timeout ? session_timeout.to_i : EC2_SESSION_TIMEOUT
  http.open_timeout = EC2_CONNECTION_TIMEOUT
end
```

lib/resolvers/ec2.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
```
IrimieBogdan, line 1: A test should be added for the resolver.

lib/resolvers/ec2.rb
oanatmaria: Done

lib/resolvers/ec2.rb
oanatmaria: Done

lib/resolvers/ec2.rb
oanatmaria: Done

lib/resolvers/ec2.rb
oanatmaria: Done

lib/resolvers/ec2.rb
oanatmaria: Done

lib/resolvers/ec2.rb
oanatmaria: These are default for all aws instances, those urls weren't changed in the last 5+ years

lib/resolvers/ec2.rb
oanatmaria: if this line is wrapped inside an ensure block, an additional check is needed in order to assure that response object is not nil.

lib/resolvers/ec2.rb
oanatmaria: I would not move this code into another method as this method is a recursive one and i think is more easy to understand if the recursive call is made from inside it.

lib/resolvers/ec2.rb
oanatmaria: Done

gimmyxd:
I suggest using https://github.com/bblimke/webmock to stub external APIs and disable external calls instead of mocking NET::HTTP.
This way, if the HTTP library will be changed in the future, the tests will not need adjustments

lib/facts/linux/ec2_userdata.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Facts
+  module Linux
+    class Ec2Userdata
+      FACT_NAME = 'ec2_userdata'
+      HYPERVISORS_HASH = { 'VMware' => 'vmware', 'VirtualBox' => 'virtualbox', 'Parallels' => 'parallels',
```
sebastian-miclea, line 7: Can this hash be moved to a different, common location? I saw that this is already defined in two other resolvers.

lib/facts/linux/ec2_userdata.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      def retrieve_from_virt_what
+        Facter::Resolvers::VirtWhat.resolve(:vm)
+      end
+
+      def check_xen
+        Facter::Resolvers::Xen.resolve(:vm)
+      end
+
+      def check_other_facts
```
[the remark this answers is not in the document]
oanatmaria: Done

lib/facts/linux/ec2_userdata.rb
```diff
@@ -0,0 +1,48 @@
...
+        Facter::ResolvedFact.new(FACT_NAME, fact_value.empty? ? nil : fact_value)
+      end
+
+      def aws?
+        virtual = retrieve_from_virt_what || check_xen || check_other_facts || check_lspci
+
+        virtual == 'kvm' || virtual =~ /xen/
+      end
+
+      def retrieve_from_virt_what
```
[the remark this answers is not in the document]
oanatmaria: Done

lib/facts/linux/ec2_userdata.rb
oanatmaria: Done

sebastian-miclea (approved):

PR #549, opened 2020-06-11
(FACT-2638) Use puppet AIO VERSION file to specify AIO version

BogdanIrimie:
Puppet `aio_agent_version` will be read from the location where puppet AIO is installed. By doing this, all facter instances installed on the system will be able to detect agent AIO version.


lib/resolvers/agent_resolver.rb
```diff
@@ -14,7 +14,7 @@ def post_resolve(fact_name)
         end
 
         def read_agent_version
-          version_file = ::File.join(ROOT_DIR, 'lib/puppet/VERSION')
+          version_file = ::File.join(ROOT_DIR, 'agent/AIO_VERSION')
           @fact_list[:aio_agent_version] = ::File.read(version_file)
```
oanatmaria, line 18: use Util::FileHelper.safe_read

gimmyxd:
I have a small concern with this approach as we will ship facter as a gem component in Puppet(that means it can be managed using puppet gem command). This means that it will be upgradable using `/opt/puppetlabas/puppet/bin gem update facter` which can install a new version of facter that will not have the agent version file.

lib/resolvers/agent_resolver.rb
IrimieBogdan: Changed the resolver name and changed to use `Util::FileHelper.safe_read`

IrimieBogdan:
Change implementation to read agent aio version from puppet installation path.

sebastian-miclea (dismissed):

lib/resolvers/windows/aio_agent_version.rb
```diff
@@ -0,0 +1,51 @@
...
+            puppet_aio_path = read_for_64_bit(reg) || read_for_32_bit(reg)
+
+            return unless puppet_aio_path
+
+            puppet_aio_version_path = File.join(puppet_aio_path, 'VERSION')
+            @fact_list[:aio_version] = Util::FileHelper.safe_read(puppet_aio_version_path, nil).chomp
+          end
+
+          def read_for_64_bit(reg)
+            reg.read('RememberedInstallDir64')[1]
```
gimmyxd, line 35: Can `reg.read('RememberedInstallDir64')` be `nil`, or if the key is not found it will just return an empty array?

lib/resolvers/windows/aio_agent_version.rb
```diff
@@ -0,0 +1,51 @@
...
+
+        class << self
+          private
+
+          def post_resolve(fact_name)
+            @fact_list.fetch(fact_name) { read_version(fact_name) }
+          end
+
+          def read_version(fact_name)
+            reg = ::Win32::Registry::HKEY_LOCAL_MACHINE.open('SOFTWARE\\Puppet Labs\\Puppet')
```
gimmyxd, line 18: What if `SOFTWARE\\Puppet Labs\\Puppet` is not present?

lib/resolvers/aio_agent_version.rb
```diff
@@ -14,8 +14,8 @@ def post_resolve(fact_name)
         end
 
         def read_agent_version
-          version_file = ::File.join(ROOT_DIR, 'lib/puppet/VERSION')
-          @fact_list[:aio_agent_version] = ::File.read(version_file)
+          aio_agent_version = Util::FileHelper.safe_read('/opt/puppetlabs/puppet/VERSION', nil).chomp
```
gimmyxd, line 17: I think @GabrielNagy mentioned that on OS X with not root, it will not be installed in `opt/puppetlabs/` but in the home of the user

GabrielNagy: I think that's mostly for Bundler usage. I don't know if you can install packages as non-root on any platform... and if you can it's definitely undocumented and shouldn't be used.

See: https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md

lib/resolvers/windows/aio_agent_version.rb
```diff
@@ -0,0 +1,51 @@
...
+            build_fact_list(reg)
+            reg.close
```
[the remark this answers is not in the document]
IrimieBogdan: Looks good, made the change

lib/resolvers/windows/aio_agent_version.rb
IrimieBogdan: Caught the exception and logged a debug message

lib/resolvers/windows/aio_agent_version.rb
IrimieBogdan: If the key does not exist, it will raise a `Win32::Registry::Error` exception.
If the key does not contain data, an array with `[1, ""]` is returned.

Added a `nil` check before `chomp` as we might not be able to read the puppet version file if there is no data in registry.

IrimieBogdan: Improved the condition for early return if no path was found in registry and added tests for this case.

gimmyxd:
@IrimieBogdan there is a small difference comparing facter 4 to facter 3
```
❯ /opt/puppetlabs/bin/facter --version
3.14.12 (commit 7b4c3bd15d90faf9978e7ef32a34e8cc91a1ca73)
❯ /opt/puppetlabs/bin/facter aio_agent_version
6.16.0.75


❯ git log -n 1
commit d775e63db4025732e60151735825762960f2ff0f (HEAD -> FACT-2638, origin/FACT-2638)
❯ bundle exec facter aio_agent_version
6.16.0.75.g625aab0b


❯ /opt/puppetlabs/bin/puppet --version
6.17.0
```

gimmyxd (dismissed):

gimmyxd (approved):

PR #556, opened 2020-06-17
Fix master

BogdanIrimie:
Currently master branch is failing because of a exception in ec2 fact. The fix catches the exception and logs a debug message.

### DIFF
--- a/lib/resolvers/ec2.rb
+++ b/lib/resolvers/ec2.rb
@@ -57,13 +57,17 @@ module Facter
           http.read_timeout = determine_session_timeout
           http.open_timeout = EC2_CONNECTION_TIMEOUT
           resp = http.get(parsed_url.path)
-          resp.code.to_i.equal?(200) ? resp.body : ''
+          response_code_valid?(resp.code) ? resp.body : ''
         rescue Net::OpenTimeout
           log.debug("#{url} timed out while trying to connect")
           ''
         rescue Errno::ENETUNREACH => e
           log.debug(e.message)
           ''
+        end
+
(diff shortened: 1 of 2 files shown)


lib/resolvers/ec2.rb
```diff
@@ -57,10 +57,17 @@ def get_data_from(url)
...
           resp = http.get(parsed_url.path)
-          resp.code.to_i.equal?(200) ? resp.body : ''
+          response_code_valid?(resp.code) ? resp.body : ''
         rescue Net::OpenTimeout
           log.debug("#{url} timed out while trying to connect")
           ''
+        rescue Errno::ENETUNREACH => e
+          log.debug(e.message)
+          ''
+        end
```
joshcooper, line 67: Ruby can raise a bunch of different exceptions such as Net::ReadTimeout, Net::WriteTimeout, OpenSSL::SSL::SSLError, and a bunch of different Errno::* exceptions. Maybe just `rescue StandardError, Timeout::Error => e` where the latter is the superclass for all of the `Net::*Timeout`?

Filipovici-Andrei: I think that rescuing StandardError is not the best thing we can do. I think that we should try to use OpenUri which is a wrapper of Net::Http and from what I saw, we can use OpenURI::HTTPError for any response different than 1xx and 2xx and exceptions.

IrimieBogdan: As this is blocking the CI, I will merge it and the suggestions can be implemented on a different branch.

Filipovici-Andrei (approved):

